PR #32, opened 2023-12-07
Vote on standard proposal

Dom-Mac:
- Cast standard proposal vote styled and working (voting power should be reviewed)
- Moved all css from emotion to modules, the emotion dependency is still used for classname concatenation, can be removed in the future


kent (changes requested):
Let's migrate the `emotion` to modules. Otherwise looks good.

kent (approved):

PR #41, opened 2023-12-19
add pagination 10 and hotfix getVotesForDelegate

ferrodri:
- paginate every 10 past votes
- hotfix getVotesForDelegate


kent (changes requested):
The Delegate view, PastVotes is not working though with this pagination.  Can you confirm and test again?

ferrodri:
Hi @kent I made this loom video where I added a 2s delay to the infinite scroll so you could see it better, this is what you meant right? https://www.loom.com/share/0a0fd54fa0c847a0905cee3371e505aa?sid=11e1a376-03c3-48dc-a1cc-94c606b92044

stepandel (approved):
LGTM

PR #42, opened 2023-12-19
Fix proposal page

Dom-Mac:
Move server action in server library


Dom-Mac:
This code doesn’t fix production issue

jjranalli (approved):

PR #43, opened 2023-12-22
Profile dropdown

Dom-Mac:
Add the profile dropdown dialog and button in the header.


kent (approved):
Fixed up the the delegates route.
Will def want to refactor the two files that are 95% the same, but good enough for now.

PR #52, opened 2024-01-12
Wojciechkulikowski/agora 1440 rounding issues in the advanced delegation logic

wojtekwtf:
change overview: https://www.loom.com/share/29afb9f91afd45a68da59f9286f2996b

additional issue for rounding up below 0.001 OP https://www.loom.com/share/8045d2938fc14620956cf1b52017e873?sid=3a18dd9c-8bf9-4a2e-9275-651afe53a52c


src/components/Dialogs/AdvancedDelegateDialog/SubdelegationRow.tsx
```diff
@@ -40,6 +42,59 @@ function SubdelegationToRow({
       ? 0
       : Math.round((allowance / availableBalanceNumber) * 100_00) / 100;
 
+  const handleAllowanceInput = (e: React.ChangeEvent<HTMLInputElement>) => {
+    function formatNumber(value: number) {
+      return Math.floor(Math.round(value * 10000) / 10) / 1000;
+    }
+
+    function getCleanInput(value: string) {
```
kent, line 50: Might be better to pull the functions out the const definition or they will keep getting defined each time.

src/lib/tokenUtils.ts
```diff
@@ -64,6 +64,29 @@ export function formatNumber(
     .join("");
 }
 
+export function formatNumberForAdvancedDelegation(amount: string | BigInt) {
+  // Advanced delegation needs a precision up to 3 decimal places,
+  // which is bit different from the formatNumber function used everywhere else and requires for max 4 significant digits
+  const number = Number(
```
kent, line 70: Given you are taking both string and BigInt you could do a check like this to normalize the input

  const amountNumber = typeof amount === 'bigint' ? Number(amount) : Number(amount);
  if (isNaN(amountNumber)) {
    throw new Error("I dont have a number here");
  }

kent (approved):
LGTM right now. Added a few comments around checks but LGTM

src/components/Dialogs/AdvancedDelegateDialog/SubdelegationRow.tsx
wojtekwtf: +1

left them inside because `formatNumber` was inside already

src/lib/tokenUtils.ts
wojtekwtf: great that you noticed it, thanks!

solving by removing the BigInt possibility, this function is used in only one place and really should only take the string

PR #53, opened 2024-01-12
AGORA-1426 As an OP Delegate, I should be able to vote on an approval proposal

Dom-Mac:
Approval proposal:
 - vote enabled
 - support new approval module


src/lib/voteUtils.ts
```diff
@@ -17,12 +17,40 @@ import { Vote } from "@/app/api/votes/vote";
...
+  switch (Number(support)) {
+    case 0:
+      return "AGAINST";
+    case 1:
+      return "FOR";
+
+    default:
+      return "ABSTAIN";
+  }
+}
```
Dom-Mac, line 44: @stepandel New approval module adds the possibility to vote for against (not needed in the ui). Support type is old approval proposal is different from new proposals. Do you confirm? Are there any other areas of the code involved ?

stepandel: just make sure we handle `optimistic` proposal type correctly

and instead of hardcoding the 114615036 directly in voteUtils, I'd add a helper function isOldModule inside `contracts.ts`
so that we can use it in other places if we ever need to
and will determine the block number depending on prod dev and OP vs other deployments

stepandel (approved):
LGTM

PR #57, opened 2024-01-12
AGORA-1455 As an OP Delegate, I should be able to vote with partial delegated voting power

stepandel:


yitongzhang:
When i tested with 0xe538f6f407937ffDEe9B2704F9096c31c64e63A8, which had 10 OP (norm) and 10.9 OP (Adv). It only triggered one trx at a time, so i had to click the button twice

stepandel:
> When i tested with 0xe538f6f407937ffDEe9B2704F9096c31c64e63A8, which had 10 OP (norm) and 10.9 OP (Adv). It only triggered one trx at a time, so i had to click the button twice.
> 
> I don't think a normal user would know to do this

@yitongzhang  There might've been an query error when pulling votes. It was resolved with merge.
This is my experience: https://www.loom.com/share/016bee4cc2b04a97b0d76deb4f958e32?sid=bfe88328-ccf9-487c-9c39-d02826b845eb

Let me know if you still having issues! A video would be nice

PR #62, opened 2024-01-16
wojeth/agora-1487 Display previously delegated balances during advanced delegation

wojtekwtf:
Solving a regression introduced in #52


kent (approved):
LGTM

PR #64, opened 2024-01-16
wojeth/agora-1464 fix non-reactive clicking issue on the delegates page

wojtekwtf:
TLDR; the clicking hitbox didn't account for the lower part of the card. Cursor pointer was defined on the whole card.


yitongzhang (approved):
LGTM!

PR #65, opened 2024-01-16
Optimisations

stepandel:
Optimizations:
- Votes query: apply proposals & voter filter directly on FDWs
- Delegate page: use `getItem` instead of `query` on read from Dynamo + parallelize loading

PS. sorting for votes query is hardcoded as I was unable to correctly parse sort order into the raw prisma string


stepandel:
Screenshot 2024-01-16 at 4 04 53 PM

PR #75, opened 2024-01-17
Frodriguezhervias/agora 1468 tablet responsive size is very broken

ferrodri:
Here is the preview link: https://agora-next-8xog6tayj-voteagora.vercel.app/

### DIFF
--- a/src/components/Hero/hero.module.scss
+++ b/src/components/Hero/hero.module.scss
@@ -30,7 +30,7 @@
     position: relative;
     height: 100%;
     width: 100%;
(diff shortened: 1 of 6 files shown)


yitongzhang:
Hey Fernando, I see you've been trying to make Agora more responsive at every breakpoint. While admirable, I don't think we're equipped to support that well right now. Instead, please do as the ticket describes and limit to only 2 sizes: mobile and web. And when a screen is between mobile and web breakpoints, show the web one version with a min width and H scroll.

Lmk if the ticket is confusing happy to describe the expected behaviour in more detail

PR #83, opened 2024-01-18
Fix bignumber parse on votes

stepandel:


stepandel:
@kent can you please check that me reverting the scientific parsing didn't break anything?
The root of the number parsing issue was coming from votes queries.
There's was a missmatch in the datatype that was returned from the DB vs what was implied when parsing the vote.

I wanna make sure that was the only use case

Generally Decimal datatype on prisma is a large floating point number and if not handled correctly, will cause parsing errors downstream

src/app/layout.tsx
```diff
@@ -13,10 +13,6 @@ BigInt.prototype.toJSON = function (): string {
 };
 
 export const metadata = {
-  metadataBase: new URL("https://vote.optimism.io"),
```
kent, line 16: You don't like my meta? ? 
Canonical is a good tag to have in there. Is there a problem with it?

kent (commented):
Did you test this on a few delegate profiles and proposals?

src/app/layout.tsx
stepandel: Oh this is unintentional. Missed that this was part of this commit as well

stepandel: sorry -- putting back

stepandel:
> Did you test this on a few delegate profiles and proposals?

Yup, votes are working fine to me. But second look doesn't hurt

stepandel:
Vote pagination doesn't seems to work. Taking another look

stepandel:
Pagination is a separate issue that happens on main as well.
Will address it in a separate PR

> Vote pagination doesn't seems to work. Taking another look

PR #93, opened 2024-01-19
wojeth/agora-1315 Needs my vote

wojtekwtf:
video: https://www.loom.com/share/0b2bc957ac3e40619cc0ef6adca00e82

note: it takes a moment to load. @yitongzhang lmk if i should place a skeleton there

### DIFF
--- a/src/app/api/proposals/getNeedsMyVoteProposals.ts
+++ b/src/app/api/proposals/getNeedsMyVoteProposals.ts
@@ -1,0 +1,53 @@
+import { notFound } from "next/navigation";
+import { paginatePrismaResult } from "@/app/lib/pagination";
+import { parseProposal } from "@/lib/proposalUtils";
+import prisma from "@/app/lib/prisma";
+import provider from "@/app/lib/provider";
+
+import "server-only";
+import { OptimismContracts } from "@/lib/contracts/contracts";
+import { getQuorumForProposal } from "../quorum/getQuorum";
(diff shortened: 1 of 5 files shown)


yitongzhang:
Hmm preview doesn't work. does it also take a long time to load if you're already signed in? A couple of questions for a potential skeleton state: 

1. We won't know how many proposals there will be in the skeleton right?
2. Is there a way to know if there will be any proposals in need my votes? would not be such a good experience to show a skeleton

Anyway – if the answer is that we can't get either, then we don't do the skeleton cause it won't be a better experience

wojtekwtf:
> Hmm preview doesn't work. does it also take a long time to load if you're already signed in?
Took around 10 seconds for me, hard to tell if it's just on dev or in prod too

> We won't know how many proposals there will be in the skeleton right?
Nope

> Is there a way to know if there will be any proposals in need my votes? would not be such a good experience to show a skeleton
Also nope

Both questions would require additional db queries

pinging @stepandel to help with the optimization

wojtekwtf:
@yitongzhang i optimized the query, the speed on average is quite good, takes around 1 second to display "Needs my vote" after loading "All proposals"

I experience the delay when i open the preview link for the first time — I don't really know why. But general performance should be fine now, please lmk what do you think.

yitongzhang:
This is sick! tysm

yitongzhang (approved):

PR #152, opened 2024-03-05
[WIP] 🛠️ Allow concurrent execution of delegate query

Flip-Liquid:
~~TODO: test locally~~ Tested on preview branch for correctness manually
~~TODO: instrument dynamo calls~~ this will come as a separate PR/ workitem
~~TODO: measure impact~~ Screenshot supplied. Seems that this has an effect from the before and after, but further optimizations are probably needed; there are queries taking ~6s as reported from our prisma logging middleware

image
^ marked improvement according to vercel speed insights

image

image

^ one particular query is taking ~6s to complete for the same delegate on prod and preview, we should tackle this one next.

---

## PR-Codex overview
This PR focuses on refactoring the `getDelegates` function to use `delegateQuery` with a `then` method instead of `await`, improving code readability and efficiency.

### Detailed summary
- Refactored `delegateQuery` usage to use `then` method for better readability
- Improved code efficiency by using `delegateQuery.then` instead of `await`

### DIFF
--- a/src/app/api/common/delegates/getDelegates.ts
+++ b/src/app/api/common/delegates/getDelegates.ts
@@ -158,7 +158,7 @@ export async function getDelegateForNamespace({
 
   const [delegate, votableSupply, delegateStatement, quorum, _isCitizen] =
     await Promise.all([
-      (await delegateQuery)?.[0] || undefined,
+      delegateQuery.then(result => result?.[0] || undefined),
       prisma[`${namespace}VotableSupply`].findFirst({}),
       getDelegateStatement({ addressOrENSName }),
       getCurrentQuorum(),


stepandel (commented):
from what I know about javascript, `await fn()` & `fn().then()` compiles into the same bite code

but I'm happy to be proven wrong!

Flip-Liquid:
> from what I know about javascript, `await fn()` & `fn().then()` compiles into the same bytecode
> 
> but I'm happy to be proven wrong!

There may be something happening under the hood that allows max concurrency here between those queries, but my naive read is that the Promise.All call won't be executed until await-completion of the delegate query

The ordering appears to be

await delegate query
await all (result of delegate query, ... Other queries)

Unless I'm missing something `await fn()` and `fn().then()` can't compile to the same bytecode, since the latter is still wrapped in a promise though, right?

stepandel:
> > from what I know about javascript, `await fn()` & `fn().then()` compiles into the same bytecode
> > but I'm happy to be proven wrong!
> 
> There may be something happening under the hood that allows max concurrency here between those queries, but my naive read is that the Promise.All call won't be executed until await-completion of the delegate query
> 
> The ordering appears to be
> 
> await delegate query await all (result of delegate query, ... Other queries)
> 
> Unless I'm missing something `await fn()` and `fn().then()` can't compile to the same bytecode, since the latter is still wrapped in a promise though, right?

Ah interesting, we've run into an issue in the past where on the client-side components the browser prevents parallel execution and queues them sequentially: https://www.reddit.com/r/nextjs/comments/13wrzbs/how_to_call_server_actions_in_parallel_and_keep/

Maybe Edge runtime on Next also has some limits

But both `await fn()` and `fn().then()` will return a promise and in Node.js runtime `Promise.all` is supped to bundle them up and execute in parallel

Flip-Liquid:
@stepandel ok if we find some time to chat synchronously today? Feel like there's something I'm missing here :)

Flip-Liquid:
@stepandel ok if we find some time to chat synchronously today? Feel like there's something I'm missing here :)

Flip-Liquid:
@stepandel here's another piece that I believe supports my understanding of what's going on 

https://www.learnwithjason.dev/blog/keep-async-await-from-blocking-execution/

Specifically the section "Two unrelated async operations"

See also the Mozilla docs which explicitly apply to Node:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

Specifically the section "Using Promise.all() with async functions"

"...in the code above, the await operator causes the async function to pause until the choice is made, and then again until the prices are fetched. We can use Promise.all to run them concurrently, so that the user doesn't have to wait for the prices to be fetched before the result is given"

Could you share some documentation for what you're saying is happening on the node side? IIRC from our meeting yesterday you mentioned that Node does some optimization which, instead of awaiting the delegateQuery, will allow it to execute concurrently with the other arguments to promise.all

stepandel:
hey @Flip-Liquid while I appreciate your persistance, I really don't think we should be arguing about JS fundamentals here.
You're providing proper references, but you're not interpreting them correctly.

Generally, you can try getting a more precise metric to measure the change that you're making. Vercel metrics can skew the measurement as there's a lot more going on under the hood. For this, you can simply log the timestamps around the execution statement https://github.com/voteagora/agora-next/blob/9d7a9f362d0d1e12ca2172ddfbdf843719854874/src/app/api/common/delegates/getDelegates.ts#L159

like so:
```
  console.log(new Date());
  const [delegate, votableSupply, delegateStatement, quorum, _isCitizen] =
    await Promise.all([
      (await delegateQuery)?.[0] || undefined,
      prisma[`${namespace}VotableSupply`].findFirst({}),
      getDelegateStatement(addressOrENSName),
      getCurrentQuorum(),
      isCitizen(address),
    ]);
  console.log(new Date());
```

And see if there're any time gain

But I'll guarantee you there won't be. Here's another test you can run:
```
await Promise.all([
      (async () => {
        console.log(new Date());
        console.log("starting fetching delegateQuery");
        return (await delegateQuery)?.[0] || undefined;
      })(),
      (() => {
        console.log(new Date());
        console.log("starting fetching votableSupply");
        return prisma[`${namespace}VotableSupply`].findFirst({});
      })(),
      (() => {
        console.log(new Date());
        console.log("starting fetching delegateStatement");
        return getDelegateStatement(addressOrENSName);
      })(),
      (() => {
        console.log(new Date());
        console.log("starting fetching quorum");
        return getCurrentQuorum();
      })(),
      (() => {
        console.log(new Date());
        console.log("starting fetching isCitizen");
        return isCitizen(address);
      })(),
    ]);
```

just as a sanity check here, it results in concurrent execution:
```
2024-03-06T20:45:03.402Z
starting fetching delegateQuery
2024-03-06T20:45:03.403Z
starting fetching votableSupply
2024-03-06T20:45:03.403Z
starting fetching delegateStatement
2024-03-06T20:45:03.403Z
starting fetching quorum
2024-03-06T20:45:03.403Z
starting fetching isCitizen
```

stepandel (approved):
sorry for being an ass especially on your first week. Nothing personal, I'm just trying to keep the team to the high standards.  My apologies, I'm actually wrong here. I appreciate your patience. This is a very convincing argument and I just learned something new!
Adding `await` before async function in `Promise.all` actually makes it behave differently.

Your typescript playground is a fantastic argument, appreciate you isolating the issue and setting up this test.

Keep rocking it! Looking forward to seeing what else you find! 💪  🔥

PR #155, opened 2024-03-05
Andrei/agora 1650 tenant abstraction contracts factory

andreitr:
Untitled (2)

The next step in our multi-tenant abstraction journey is complete. The most significant change this time around is that I've replaced references to OptimismContracts with tenant-scoped contract definitions via `tenant.contracts.governor`, etc. Also, I've swapped out OP-specific implementations (e.g., `OptimismGovernorInterface`) for generic interfaces like `IGovernorContract`, `ITokenContract`. This shift allows us to start programming to interfaces rather than specific implementations.

---

## PR-Codex overview
This PR updates imports related to `Tenant` to include a subdirectory path. 

### Detailed summary
- Updated imports of `Tenant` to include a subdirectory path in multiple files
- Added new types/interfaces related to contracts
- Refactored `TenantTokenFactory` class
- Adjusted usage of `Tenant` instance and contracts in various functions

> The following files were skipped due to too many changes: `src/app/lib/hooks/useIsOpManager.ts`, `src/app/delegates/actions.ts`, `src/components/Proposals/ProposalPage/OPProposalPage/OpManagerDeleteProposal.tsx`, `src/app/api/common/quorum/getQuorum.ts`, `src/app/api/common/proposals/getProposals.ts`, `src/lib/alligatorUtils.ts`, `src/components/Dialogs/AdvancedDelegateDialog/AdvancedDelegateDialog.tsx`, `src/app/lib/hooks/useIsAdvancedUser.ts`, `src/lib/tokenUtils.ts`, `src/hooks/useAdvancedVoting.tsx`, `src/components/Admin/ProposalType.tsx`, `src/app/api/common/delegates/getDelegates.ts`, `src/components/Dialogs/DelegateDialog/DelegateDialog.tsx`, `src/components/Dialogs/AdvancedDelegateDialog/useAdvancedDelegation.tsx`, `src/lib/tenant/tenant.ts`, `src/components/Proposals/ProposalPage/OPProposalApprovalPage/OptionResultsPanel/OptionResultsPanel.tsx`, `src/components/Admin/GovernorSettings.tsx`, `src/components/Proposals/ProposalCreation/SubmitButton.tsx`, `src/app/api/common/delegations/getDelegations.ts`, `src/app/api/common/voting-power/getVotingPower.ts`, `src/lib/tenant/tenantContractFactory.ts`, `src/lib/contracts/contracts.ts`


src/app/lib/hooks/useIsAdvancedUser.ts
```diff
@@ -66,10 +67,10 @@ const useIsAdvancedUser = () => {
      * PROD: only allowlist
      * TEST: more than 1 token or allowlist
      */
-    onSuccess: (balance) => {
+    onSuccess: (balance: bigint) => {
```
andreitr, line 70: cc @stepandel @kent 

Once I replaced contract-specific classes (e.g., `OptimismToken`) with generic interfaces (e.g., `IContractToken`), I realized that return types were not being automatically inferred, hence the type casting like `balance: bigint`. If you guys know a way to infer types from interfaces, please let me know ;)

stepandel: the `onSuccess()` type is inferred from the `abi` type. Since it's `any`, it won't be able detect the types for either `args` or `onSuccess()`

You can technically write an `AbiType` for each `ContractType` if it's worth the effort.

One approach could be have a generic ABI for Token & governor contracts, maybe even alligator, which is a stripped down ABI with only a few methods that are actually in use: https://github.com/voteagora/agora-next/blob/aebddb3332b02ebeb5920510c72bcb66176decfe/src/lib/tenant/tenantContractFactory.ts#L42

src/app/api/common/quorum/getQuorum.ts
```diff
@@ -1,15 +1,14 @@
 import provider from "@/app/lib/provider";
 import prisma from "@/app/lib/prisma";
-import { contracts } from "@/lib/contracts/contracts";
 import { ProposalPayload } from "../proposals/proposal";
-import Tenant from "@/lib/tenant";
+import Tenant from "@/lib/tenant/tenant";
 
 export async function getQuorumForProposal(proposal: ProposalPayload) {
-  const { namespace } = Tenant.getInstance();
+  const { namespace, contracts } = Tenant.getInstance();
```
Flip-Liquid, line 7: @andreitr,

if these are being read from the process env, then we're not expecting these to change over the process lifetime. we could potentially DRY this by loading the ns and contracts once per process as well. 

this way, we can have the tenant namespace and contracts "resolved" before we enter these queries. looking forward to "dynamic" tenant resolution (i.e. retrieving details at runtime rather than from process env) we would also have a clearer separation of concerns between tenant resolution and query execution, which would allow for easier instrumentation of each of these.

but this can also be an incremental change, just a thought for the future

andreitr: @Flip-Liquid good catch, I am sorta getting around this issue using a singleton, which should initialize the contracts and other configs, only once. The performance hit should be minimal unless nextjs does something funky with back-end vs front-end initialization.

src/app/lib/hooks/useIsAdvancedUser.ts
andreitr: @stepandel I see. I think building a generic abi makes sense. I can add it to my todo list for the `ens` implementation.

ferrodri:
Code looks great to me @andreitr, once the indexer is back I will tell you if I find any errors during QA! (probably not)

PR #163, opened 2024-03-11
Prisma client instrumentation + common logging library

Flip-Liquid:
Small PR to:
- remove use of deprecated `$use` client method on PrismaClient in favor of client extension
  - seems to fix a noisy logging bug I've seen where the `$use` callback is executed many times for the same query 
- introduce a logging library used in above, and which we can expand for greater observability 
- instrument dynamoDB access

---

## PR-Codex overview
The focus of this PR is to add logging functionality to Prisma queries for performance monitoring.

### Detailed summary
- Added `time_this` function in `logging.ts` for timing functions and logging
- Implemented logging middleware in `prisma.ts` for monitoring query performance
- Modified `prisma.ts` to use `time_this` for tracking query execution time


src/app/lib/prisma.ts
```diff
@@ -1,35 +1,62 @@
...
+      async $executeRawUnsafe({ args, query, operation }) {
+        return await execRaw(query, args, operation);
+      },
+    },
+  });
+};
+
 if (process.env.NODE_ENV === "production") {
-  prisma = new PrismaClient();
+  prisma = makePrismaClient() as PrismaClient;
```
andreitr, line 52: Does this also run in prod?

andreitr (approved):
Left a nonblocking comment. LGTM 🚢

src/app/lib/prisma.ts
Flip-Liquid: I think that line would. I didn't trace out what was going on with the conditionals on the client instantiation, so I kept the environment checks the same.

PR #166, opened 2024-03-12
Ferrodri/agora 1727 fix loading on voting cast not executing

ferrodri:
I added react query, just for client queries, should not be used for server revalidation since we have other methods available from nextjs.

Here is the preview link: https://agora-next-aih93ufbb-voteagora.vercel.app/proposals/24501583086785187277158752896455217063239294778481708626402919413853617969721
Here is an older preview link where you can test voting is not working: https://agora-next-j658wjc9e-voteagora.vercel.app/proposals/24501583086785187277158752896455217063239294778481708626402919413853617969721

Here is a video with the explanation: https://www.loom.com/share/ad960880ad8244fb957139b2a0479937?sid=df6d2179-4e0e-4763-a2f9-3b11eb2378d6

I am also solving a rangeError that was occurring at alligatorUtils, here is a screenshot: 
math trunc-error

LGTM!

---

## PR-Codex overview
The focus of this PR is to integrate `@tanstack/react-query` for data fetching and state management in the app.

### Detailed summary
- Added `@tanstack/react-query` for data fetching and state management
- Updated `cast_vote_container` styling in `CastVoteInput`
- Refactored `useConnectedDelegate` hook to use `useQuery` for fetching data

### DIFF
--- a/src/hooks/useConnectedDelegate.ts
+++ b/src/hooks/useConnectedDelegate.ts
@@ -4,6 +4,7 @@ import { Delegate } from "@/app/api/common/delegates/delegate";
 import { useState, useCallback, useEffect } from "react";
 import { useConnectButtonContext } from "@/contexts/ConnectButtonContext";
 import { fetchDelegate } from "@/app/delegates/actions";
+import { useQuery } from "@tanstack/react-query";
 
 function timeout(ms: number) {
   return new Promise((resolve) => setTimeout(resolve, ms));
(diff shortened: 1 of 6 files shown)


src/hooks/useConnectedDelegate.ts
```diff
@@ -21,6 +22,11 @@ const useConnectedDelegate = () => {
   const [balance, setBalance] = useState<bigint | null>(null);
   const [retries, setRetries] = useState<number>(0);
   const [lastVotingPower, setLastVotingPower] = useState<string | null>(null);
+  useQuery({
```
andreitr, line 25: I'm generally conservative when it comes to adding new library dependencies, but in this case, replacing `useEffect` with `useQuery` makes a lot of sense.

src/lib/alligatorUtils.ts
```diff
@@ -122,8 +122,8 @@ export async function getTotalVotableAllowance({
 
   const totalAllowance =
     (allowances.reduce((a, b) => a + b, 0n) *
-      BigInt((1 - subdelegatedShare) * 100000)) /
-      100000n -
+      BigInt(Math.trunc((1 - subdelegatedShare) * 100000))) /
```
andreitr, line 125: I don't understand how is this different from the original code?

ferrodri: Math.trunc removes the decimals so converts 24931.000000000004 into 24931 so there is no error when converting into a bigint
math trunc-error

andreitr (approved):
LGTM 🚢

PR #171, opened 2024-03-15
Ferrodri/agora 1708 fix "when entering a value for delegation i am getting a negative number"

ferrodri:
Here is the preview link: https://agora-next-5acdmtr2s-voteagora.vercel.app/delegates/lefteris.eth, to reproduce this exact use case you can log in with agora manager throwaway wallet 0xe538f6f407937ffDEe9B2704F9096c31c64e63A8

Here is the error
Screenshot 2024-03-15 at 18 13 40

Here is the solution, @yitongzhang it would be great if you could confirm if "total delegatable votes reached" is a good copy for this use case.
Screenshot 2024-03-15 at 19 25 44

Here is a loom video in case you guys want to watch my explanation for further understanding: https://www.loom.com/share/a75778809bc6415194c997ec217ad64f?sid=d3650e2a-2b17-4332-a4af-fad40b9d1fdc

---

## PR-Codex overview
The focus of this PR is to enhance the user interface and functionality of the Advanced Delegation Dialog component. 

### Detailed summary
- Improved styling for subdelegation rows
- Added overflow delegation warning
- Enhanced user input experience for allowances
- Updated state management for overflow delegation
- Improved button handling for delegation actions


yitongzhang:
Ok! so it seems like there are two things needed here

1. preventing users from delegating more than they have on the UI.

2. handle the case where they've delegated more than is possible

For 1. this should already be not possible right? The only way would be to hit the contract directly? If you can confirm that this is already not possible, then nothing else needed.

For 2. seems like we need to:

a) show that you've delegated more than your total allowed amount 
b) prevent you from delegating more 
c) ask you to undelegate the excess delegation

Something like this?
image

ferrodri:
> Ok! so it seems like there are two things needed here
> 
> 1. preventing users from delegating more than they have on the UI.
> 2. handle the case where they've delegated more than is possible
> 
> For 1. this should already be not possible right? The only way would be to hit the contract directly? If you can confirm that this is already not possible, then nothing else needed.
> 
> For 2. seems like we need to:
> 
> a) show that you've delegated more than your total allowed amount b) prevent you from delegating more c) ask you to undelegate the excess delegation
> 
> Something like this? (this is assuming you've already delegated 110 OP to yitong, and are trying to delegate more to polynya) image

Point 1 confirmed, point 2 makes a lot of sense, much better UI-UX :)

andreitr (approved):
Fixed confirmed!

ferrodri:
> Fixed confirmed!

Still have to do a couple changes, will keep you updated!

ferrodri:
Here is the preview link @yitongzhang @andreitr https://agora-next-fjgyyjx38-voteagora.vercel.app/

Screenshot here: 
Screenshot 2024-03-21 at 21 39 43

yitongzhang:
UI lgtm ty!

yitongzhang (approved):

andreitr (approved):
🚢

PR #187, opened 2024-03-21
[1766, 1768] Accept and validate API query parameters, Extend pagination for limit/offset semantics

Flip-Liquid:
## PR-Codex overview
This PR adds API routes and validators for delegates and proposals, implements pagination logic, and updates delegate data fetching.

### Detailed summary
- Added API routes for delegates and proposals
- Implemented pagination logic for delegate data
- Updated delegate data fetching with new types and functions
- Added validators for input validation in API routes

> The following files were skipped due to too many changes: `src/app/api/common/delegates/getDelegates.ts`


src/app/api/common/delegates/getDelegates.ts
```diff
@@ -13,13 +18,102 @@ import {
...
+  };
+
+  const { meta, data: delegates } = await paginateResultEx(
+    paginatedQuery,
+    pagination
+  );
+  const _delegates = await Promise.all(
+    delegates.map(async (delegate: DelegatesGetPayload) => {
+      return {
+        citizen: await fetchIsCitizen(delegate.delegate),
```
stepandel, line 87: I think `isCitizen` check one can be incorporated into the main `apiDelegatesQuery` since the data is coming from the same DB

stepandel: Also, thinking for the future here, delegate statements are going to live in the same DB as well. So, ideally, we can join all this data in the same query

stepandel (commented):
love the zod validators!

src/app/api/common/delegates/getDelegates.ts
Flip-Liquid: done, touched that up

src/app/lib/pagination.ts
```diff
@@ -7,33 +7,73 @@ type PaginatedResult<T> = {
...
-  const theData = data.slice(0, pageSize) as T;
+  const has_next = data.length > params.limit;
+  const theData = data.slice(0, params.limit) as T;
 
   return {
     meta: {
-      currentPage: page,
-      pageSize,
-      hasNextPage,
+      has_next: has_next,
```
andreitr, line 73: is there a reason to change the meta format from hasNextPage to has_next etc?

andreitr (commented):
Looks good. Left a small comment. We

src/app/lib/pagination.ts
Flip-Liquid: The pagination semantics should not change for existing calls to PaginateResult. The semantics are slightly different for our external API (which is using limit/offset as opposed to pageSize + page index). 

The reason for limit / offset for our API is that it seems to be a slightly-more-often-used semantic compared to page size, so was using that to keep dev integration as simple as possible.

completely open to either snake/camel case for the pagination metadata. I really have no strong opinion either way, we just need to be consistent once we start exposing it.

stepandel (approved):
LGTM!

andreitr (commented):
LGTM

PR #196, opened 2024-03-22
Secure spanshot route

stepandel:
## PR-Codex overview
This PR enhances the snapshot route by implementing authorization with a secret key check and logging the number of updated proposals.

### Detailed summary
- Added authorization check using a secret key
- Implemented logging for the number of updated proposals
- Modified `GET` function to accept a `NextRequest` parameter

### DIFF
--- a/src/app/api/snapshot/route.ts
+++ b/src/app/api/snapshot/route.ts
@@ -84,6 +84,10 @@ export async function GET(request: NextRequest) {
         });
       }
 
+      console.log(
+        `Snapshot cron executed. Updated: ${proposals.length} proposals`
+      );
+
       return new Response(`Updated: ${proposals.length} proposals`, {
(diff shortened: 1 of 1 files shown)


src/app/api/snapshot/route.ts
```diff
@@ -75,6 +84,10 @@ export async function GET() {
         });
       }
 
+      console.log(
+        `Snapshot cron executed. Updated: ${proposals.length} proposals`
+      );
```
stepandel, line 89: @Flip-Liquid wondering if this is the right way to log this?

Flip-Liquid: For now, this will work. but likely, we'll want to use a common cron logger or some such logger with common attributes.

Flip-Liquid (approved):
looks great, thanks for adding

PR #199, opened 2024-03-23
[1772, 1796] Authenticate calls to /delegates route

Flip-Liquid:
## PR-Codex overview
This PR focuses on enhancing API user authentication and key generation. It includes middleware updates, API key generation script, and Prisma schema changes.

### Detailed summary
- Added new middleware for API key validation
- Implemented API key generation script
- Updated Prisma schema for API users
- Modified authentication logic for API users
- Updated dependencies including `uuid` and `ts-node`

> The following files were skipped due to too many changes: `yarn.lock`


src/app/lib/middleware/auth.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  let authResponse: AuthResponse = hasApiKey(request);
+
+  const token = request.headers.get("authorization");
+
+  // TODO: caching logic, rate limiting
+  const user = await prisma.api_user.findFirst({
+    where: {
+      api_key: token,
```
stepandel, line 49: Best practice is to hash these

stepandel (commented):
If we're storing API keys in the DB, we should be hashing them. But considering the middleware constrain (no prisma in edge runtime), wondering if using an encryption based key like JWT might be a better approach there?

src/app/lib/middleware/auth.ts
Flip-Liquid: true true, miss on my part

Flip-Liquid: Addressed this, we're now comparing the hash

stepandel (approved):
LGTM!

PR #200, opened 2024-03-25
[1777] Fix voteHoverCard -> fetchDelegate server action not working

ferrodri:
Hi @andreitr @Flip-Liquid @stepandel would love to get your feedback on this draft PR.

When hovering over every voter, with standard server actions only one action at a time is being processed, which looked like it was canceling the requests when hovering over multiple voters at the same time since the first request takes >2s to resolve and all the other requests during this 2s are effectively not being done.

VoteHoverCard was never resolving as you can see on the screenshot, you can try to hover over multiple voters here https://vote.optimism.io/proposals/47253113366919812831791422571513347073374828501432502648295761953879525315523 to check data is never resolved except for the first hover.
Screenshot 2024-03-25 at 13 17 20

I was looking at react "use server" docs: https://react.dev/reference/react/use-server#serializable-parameters-and-return-values and more concretely saw this:
>  Server Actions are designed for mutations that update server-side state; **they are not recommended for data fetching**. Accordingly, **frameworks implementing Server Actions typically process one action at a time and do not have a way to cache the return value.**

So I decided to give it a shot and do a random fetch to a random GET to see if I was not having this problem about requests not being done. I thought it made sense, following the docs, to just do a fetch GET request to our API endpoint to retrieve some data. Now it works perfectly fine and the bug is solved. In the future, client requests to the backend that do data fetching should probably be done this way, due to all the errors in this sense we have been having these past days. But this is something we can test in the upcoming weeks. 

Lastly, here is a video demonstrating this statement: https://www.loom.com/share/b7b5bbc6423147078d415ca01eadfda6?sid=95c978d0-a0b7-4a7b-ba46-f8938b62ae9f

---

## PR-Codex overview
This PR focuses on restructuring API routes, updating authentication middleware, and improving API request handling. 

### Detailed summary
- Deleted several unused files related to proposals and delegates
- Updated authentication logic in `auth.ts`
- Added a new `useFetchDelegate` hook for fetching delegate data
- Updated API request handling in `agoraAPI.js`
- Restructured API routes in `next.config.js`
- Added a new API route for fetching delegates in `route.ts`
- Updated component to use `useFetchDelegate` hook in `VoterHoverCard.tsx`


src/app/api/v1/delegates/[addressOrENSName]/route.ts
```diff
@@ -0,0 +1,13 @@
+import { NextResponse, type NextRequest } from "next/server";
+import { fetchDelegate } from "@/app/api/common/delegates/getDelegates";
+
+export async function GET(request: NextRequest) {
+    try {
+        const addressOrENSName = request.nextUrl.pathname.split('/')[4];
+        const delegate = await fetchDelegate(addressOrENSName);
```
Comment (ferrodri) [inline src/app/api/v1/delegates/[addressOrENSName]/route.ts:7]: @Flip-Liquid here we are importing fetchDelegate which is caching getDelegate, however, since we will use fetch on the frontend it will already cache the result of this GET request by default, should I leave it as it is, or expose getDelegate without cache just for this API route?

Comment (Flip-Liquid) [reply src/app/api/v1/delegates/[addressOrENSName]/route.ts:7]: The /api/v1 routes are going to be authenticated, and primarily intended for use externally. I wouldn't recommend this pattern, since we'd basically need to embed an API Key on the client to run this fetch. 

Can we make the hover a server component and pre fetch the info for the VoterHoverCards being displayed while on the server? I feel like that might be more performant as well.

Flip-Liquid:
@ferrodri @andreitr leaving a few more thoughts here following up from yesterday, overall good with the approach we discussed

I think it would still behoove us to push as much rendering/data fetching onto the server as possible, but as we discussed, it does not quite fit our use for the hover cards (infeasible to prefetch ~1000s of delegates)

for client fetching data, we can move ahead with using the same API surface as we'll use for external developers.

@ferrodri can you merge this to add in the authentication lib? we can then integrate key validation into the route you have above.

Also, the below might be a useful wrapper you can co-opt for use in the FE:
agora-next/src/app/lib/agoraAPI.js

ferrodri:
> @ferrodri @andreitr leaving a few more thoughts here following up from yesterday, overall good with the approach we discussed
> 
> I think it would still behoove us to push as much rendering/data fetching onto the server as possible, but as we discussed, it does not quite fit our use for the hover cards (infeasible to prefetch ~1000s of delegates)
> 
> for client fetching data, we can move ahead with using the same API surface as we'll use for external developers.
> 
> @ferrodri can you merge this to add in the authentication lib? we can then integrate key validation into the route you have above.
> 
> Also, the below might be a useful wrapper you can co-opt for use in the FE: agora-next/src/app/lib/agoraAPI.js

Great! Could you approve, just converted it to a ready for review PR. Here is the preview link: https://agora-next-pammpl1rp-voteagora.vercel.app/

Flip-Liquid (changes requested):
two points of feedback:

1) let's keep the API routes as plural for both /delegates and /delegates/{address}
2) can you add authentication to the /delegates/[address] route? similar to the /delegates route

ferrodri:
> two points of feedback:
> 
> 1. let's keep the API routes as plural for both /delegates and /delegates/{address}
> 2. can you add authentication to the /delegates/[address] route? similar to the /delegates route

Changes done, LGTM :)

Flip-Liquid:
@ferrodri, looks good, but I think you're still deleting src/app/api/v1/delegates/[addressorENSName]/route.ts in the PR

Flip-Liquid (approved):

PR #201, opened 2024-03-25
Fixes 1755, 1756, 1653

ferrodri:
Fixes the following 3 linear issues:
- https://linear.app/agora-app/issue/AGORA-1756/in-an-approval-proposal-if-options-have-no-transactions-nothing-is
- https://linear.app/agora-app/issue/AGORA-1755/in-a-top-choices-approval-proposal-top-options-are-not-shown-correctly
- https://linear.app/agora-app/issue/AGORA-1653/transaction-description-shows-incorrect-value

### DIFF
--- a/src/components/Proposals/ProposalPage/ApprovedTransactions/OptionDescription.jsx
+++ b/src/components/Proposals/ProposalPage/ApprovedTransactions/OptionDescription.jsx
@@ -1,13 +1,19 @@
 import TokenAmountDisplay from "@/components/shared/TokenAmountDisplay";
 import { shortAddress } from "@/lib/utils";
+import Tenant from "@/lib/tenant/tenant";
 
 export default function OptionDescription({ description, value, target }) {
+  const { token } = Tenant.current();
+
   return (
     <span>
-      {/* TODO: Warning – this dangerously assumes all tokens are OP. For now, ok, but won't always be the case */}
       {"//"} {description} requesting{" "}
-      <TokenAmountDisplay amount={value} decimals={18} currency="OP" /> transfer
-      to {shortAddress(target)}
+      <TokenAmountDisplay
+        amount={value}
+        decimals={token.decimals}
+        currency={token.symbol}
+      />{" "}
+      transfer to {shortAddress(target)}
     </span>
   );
 }
--- a/src/components/Proposals/ProposalPage/ApprovedTransactions/ApprovedTransactions.jsx
+++ b/src/components/Proposals/ProposalPage/ApprovedTransactions/ApprovedTransactions.jsx
@@ -40,7 +40,7 @@ export default function ApprovedTransactions({ proposalData }) {
                 <p className="font-mono text-xs font-medium leading-4 text-gray-af">
                   <OptionDescription
                     description={option.description}
-                    value={option.values[0]}
+                    value={option.budgetTokensSpent}
                     target={option.targets[0]}
                   />
                 </p>
--- a/src/components/Proposals/ProposalPage/ApprovedTransactions/ApprovedTransactions.jsx
+++ b/src/components/Proposals/ProposalPage/ApprovedTransactions/ApprovedTransactions.jsx
@@ -35,26 +35,28 @@ export default function ApprovedTransactions({ proposalData }) {
                 ? `{ value: ${formatEther(option.values[0])} ETH }`
                 : undefined;
 
-            if (option.values.length > 0) {
-              return (
-                <div key={index}>
-                  <p className="font-mono text-xs font-medium leading-4 text-gray-af">
-                    <OptionDescription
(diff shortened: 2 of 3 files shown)


src/components/Proposals/ProposalPage/ApprovedTransactions/OptionDescription.jsx
```diff
@@ -1,13 +1,19 @@
...
 export default function OptionDescription({ description, value, target }) {
+  const { token } = Tenant.current();
+
   return (
     <span>
-      {/* TODO: Warning – this dangerously assumes all tokens are OP. For now, ok, but won't always be the case */}
       {"//"} {description} requesting{" "}
-      <TokenAmountDisplay amount={value} decimals={18} currency="OP" /> transfer
-      to {shortAddress(target)}
+      <TokenAmountDisplay
```
andreitr, line 11: Nice! Another option is to move all of the tenant token stuff into the TokenAmountDisplay component.

andreitr (approved):
LGTM

src/components/Proposals/ProposalPage/ApprovedTransactions/OptionDescription.jsx
ferrodri: Went for the refactor in this commit: https://github.com/voteagora/agora-next/commit/4c88c9e5f2a6146fbdbdb7a6b308a8e20be7c3ca lmk what do you think :)

src/components/shared/TokenAmountDisplay.tsx
```diff
@@ -1,16 +1,18 @@
 import { formatNumber } from "@/lib/utils";
 import { BigNumberish } from "ethers";
 import React, { useMemo } from "react";
+import Tenant from "@/lib/tenant/tenant";
+const { token } = Tenant.current();
```
andreitr, line 5: This is interesting. You made the token a class-level variable instead of the instance-level. I have never done things this way but I guess it works 🤷‍♂️

src/components/shared/TokenAmountDisplay.tsx
```diff
@@ -1,16 +1,18 @@
...
   maximumSignificantDigits = 2,
 }: {
   amount: BigNumberish;
-  decimals: number;
-  currency: string;
+  decimals?: number;
+  currency?: string;
   maximumSignificantDigits?: number;
 }) {
   const formattedNumber = useMemo(() => {
```
andreitr, line 18: Oh I see. My preference would be to only keep the `amount` in the props and make decimals and symbol instance variables. But this also works.

src/lib/utils.ts
```diff
@@ -79,12 +78,17 @@ export function formatNumber(
   return numberFormat.format(standardUnitAmount);
 }
 
-export function TokenAmountDisplay(
-  amount: string | BigNumberish,
-  decimals: number,
-  currency: string,
+export function TokenAmountDisplay({
```
andreitr, line 81: Can we combine these methods? Maybe a static inside `TokenAmountDisplay`?

andreitr (approved):
Left a few non-blocking comments. I will leave up to @ferrodri to decide whether to make the suggested changes.

src/components/shared/TokenAmountDisplay.tsx
ferrodri: I thought this is what you meant, lmk otherwise. I think this is great since reduces a lot of duplicate code while still giving some flexibility :)

src/lib/utils.ts
ferrodri: Do you mean something like this?
`import { useMemo } from "react";
import { formatNumber } from "@/lib/utils";
import { BigNumberish } from "ethers";
const { token } = Tenant.current();

type TokenAmount = {
  amount: BigNumberish;
  decimals?: number;
  currency?: string;
  maximumSignificantDigits?: number;
};

export function formatTokenAmount({
  amount,
  decimals = token.decimals,
  currency = token.symbol,
  maximumSignificantDigits = 2,
}: TokenAmount) {
  const formattedNumber = formatNumber(
    amount,
    decimals,
    maximumSignificantDigits
  );
  return `${formattedNumber} ${currency}`;
}

export function TokenAmountDisplay({
  amount,
  decimals = token.decimals,
  currency = token.symbol,
  maximumSignificantDigits = 2,
}: TokenAmount) {
  const formattedAmount = useMemo(
    () =>
      formatTokenAmount({
        amount,
        decimals,
        currency,
        maximumSignificantDigits,
      }),
    [amount, decimals, currency, maximumSignificantDigits]
  );

  return <span>{formattedAmount}</span>;
}
`

ferrodri: I am going to merge, since I believe this would not make a big difference

PR #216, opened 2024-04-03
Ether.fi delegate statement + Code of Conduct toggle 

andreitr:
- implemented Ether.fi delegate statement creation 
- removed code of conduct toggle for Ether.fi tenant

### DIFF
--- a/src/lib/serverVerifyMessage.ts
+++ b/src/lib/serverVerifyMessage.ts
@@ -1,27 +1,53 @@
 import "server-only";
 
-import { createPublicClient, http } from 'viem'
+import { createPublicClient, http } from "viem";
 import { mainnet, optimism } from "viem/chains";
-import { type DaoSlug } from "@prisma/client";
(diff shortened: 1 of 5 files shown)


src/lib/serverVerifyMessage.ts
```diff
@@ -1,27 +1,53 @@
...
-        signature
-    });
-}
\ No newline at end of file
+  return await publicClient.verifyMessage({
+    address,
+    message,
+    signature,
+  });
+}
```
stepandel, line 53: I know this logic was there already but we don't need to create a public client, we should have access to provider on the backend already

stepandel: we can use [`ethers.verifyMessage` ](https://docs.ethers.org/v6/api/hashing/#verifyMessage)

ferrodri (approved):
Apart from @stepandel fix I think it looks good,

PR #233, opened 2024-04-11
[1884, 1854] Move connected delegate server action to client

ferrodri:
Here is the preview link: https://agora-next-iqh1yhlzs-voteagora.vercel.app/

Leave it as a draft PR in case you want to change the route URL @Flip-Liquid

---

## PR-Codex overview
The focus of this PR is to refactor the `useConnectedDelegate` hook to optimize data fetching and implement retry logic for delegate voting power updates.

### Detailed summary
- Removed `fetchConnectedDelegate` function
- Refactored data fetching using `api.get`
- Implemented retry logic with exponential backoff for delegate voting power updates

### DIFF
--- a/src/hooks/useConnectedDelegate.ts
+++ b/src/hooks/useConnectedDelegate.ts
@@ -6,6 +6,12 @@ import { useQuery } from "@tanstack/react-query";
 import AgoraAPI from "@/app/lib/agoraAPI";
 import Tenant from "@/lib/tenant/tenant";
 
+/**
+ * Define maximum number of retries, max retries 10 means 180 seconds waiting in total (advanced delegation voting power
+ * takes around 120 seconds to update)
+ */
+const MAX_RETRIES = 10;
+
 function timeout(ms: number) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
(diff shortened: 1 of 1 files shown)


andreitr (changes requested):
I am a bit concerned about uncontrolled retry attempts. The retry logic in the code does not have a limit on the number of retries. This could potentially lead to an infinite loop. I will send you a code sample in Slack.

src/hooks/useConnectedDelegate.ts
```diff
@@ -24,35 +30,54 @@ const useConnectedDelegate = () => {
...
          */
         if (
           delegatee.votingPower === refetchDelegate.prevVotingPowerDelegatee
         ) {
-          await timeout(2000);
-          const _retries = retries + 1;
-          setRetries(_retries);
+          // Check if maximum retries has been reached
+          if (retries < MAX_RETRIES) {
+            // Implement exponential backoff
```
Flip-Liquid, line 57: @ferrodri nice work on the retries. can we refac this into something reusable for other components? we'll probably need safe retry logic in other places in our app

src/hooks/useConnectedDelegate.ts
```diff
@@ -24,35 +30,54 @@ const useConnectedDelegate = () => {
...
         return { delegate, advancedDelegators, balance };
       } else if (refetchDelegate) {
         // When refetchDelegate is true, if last voting power is equal to actual it means indexer has not indexed the
         // new voting power
         if (delegate.votingPower === lastVotingPower) {
-          await timeout(2000);
-          const _retries = retries + 1;
-          setRetries(_retries);
+          // Check if maximum retries has been reached
+          if (retries < MAX_RETRIES) {
```
Flip-Liquid, line 72: nice work. can we DRY this?

Flip-Liquid:
@ferrodri looks like this got merged before I had a chance to review here.

Have we considered instead rendering this on the server? I'd rather prefer that approach, since this was not one of the public routes we were planning for (but can definitely design and talk about)

Flip-Liquid:
@stepandel & I also talked about spec-driven API development. since we're adding each of these routes to our public API surface, anyone can go out and access them. For public routes at least, I think we're going to prefer a dev process where we propose, review, and apply changes to our formal API specification, use this specification to autogenerate typescript types, documentation, and potentially client code, and then implement the routes for consumption. 

let's all talk @ferrodri @andreitr @stepandel about a good approach we can take here for adding new endpoints.

ferrodri:
> @ferrodri looks like this got merged before I had a chance to review here.
> 
> Have we considered instead rendering this on the server? I'd rather prefer that approach, since this was not one of the public routes we were planning for (but can definitely design and talk about)

Hi @Flip-Liquid this was already implemented on the server and live during the past month. To be honest this is a minor improvement, but can improve performance if there is another server action being run. I did this improvement since Yitong and other users reported a slow loading of their profile info. Furthermore, I think later on we should optimize these queries on the backend since it still looks slow. But maybe moving it to the server back again is not a good idea?

src/hooks/useConnectedDelegate.ts
ferrodri: will implement this in another PR

src/hooks/useConnectedDelegate.ts
ferrodri: will implement this in another PR

PR #234, opened 2024-04-15
[AGORA-1926] RPGF OAS

Flip-Liquid:
## PR-Codex overview
The focus of this PR is enhancing the OpenAPI spec for the Agora platform, introducing new components like RetroFunding rounds, projects, and ballots.

### Detailed summary
- Updated description and security schemes
- Added new schemas: AuthToken, SIWEVerificationBody, VotingToken, Comment
- Introduced RetroFunding components: Round, Event, ImpactMetric, Project, Ballot, Funding
- Defined parameters for API paths
- Added authentication endpoints for SIWE
- Improved error responses

> The following files were skipped due to too many changes: `spec/oas_v1.yaml`


spec/oas_v1.yaml
```diff
@@ -1035,10 +1426,317 @@ paths:
...
+              schema: 
+                $ref: "#/components/schemas/RPGFRound"
+        '400':
+          description: Bad Request
+        '401': 
+          description: Unauthorized
+        '500':
+          description: Internal Server Error
+  /rpgf/rounds/{roundId}/ballots:
+  # Should this be paginated? How many ballots per round are expected?
```
stepandel, line 1466: yes good call, we gonna have a couple of 100s of ballots

stepandel: btw is there a way to hide the admin only endpoints from public? Not sure if it's possible to have it in the yaml but hide it from the public docs?

spec/oas_v1.yaml
```diff
@@ -466,6 +493,269 @@ components:
...
+          format: date-time
+        details:
+          type: string
+    RPGFBallotVote:
+      summary: A vote for an RPGF ballot
+      description: >
+        A vote for an RPGF ballot, including information about the vote.
+      type: object
+      properties:
+        projectId:
```
stepandel, line 705: should this be metricId?

spec/oas_v1.yaml
Flip-Liquid: I can check if there are any OAS level primitives which might help. we could always have particular endpoints documented in a private repository. what's the concern with having these documented, if inaccessible otherwise?

spec/oas_v1.yaml
```diff
@@ -466,6 +518,266 @@ components:
...
+          type: string
+    RetroFundingBallotAllocation:
+      summary: A vote for an RetroFunding ballot
+      description: >
+        A vote for an RetroFunding ballot, including information about the vote.
+      type: object
+      properties:
+        metricId:
+          type: integer
+        categoryId:
```
stepandel, line 734: a bit nit picking here but we can probably drop `categoryId` -- don't really see how it can be used here

stepandel (approved):
Looking great! ✅ 🚀

PR #244, opened 2024-04-22
💻 add mocked route for ballots

Flip-Liquid:
## PR-Codex overview
The focus of this PR is to update the RetroFunding API endpoints related to fetching and submitting ballots. It includes adding routes for fetching and submitting specific ballots based on round ID and caster address/ENS.

### Detailed summary
- Added routes for fetching and submitting specific ballots based on round ID and caster address/ENS
- Updated API functions to fetch individual ballots based on round ID and caster address/ENS
- Renamed parameters in OpenAPI spec to reflect the caster address instead of ballot address/ENS


src/app/api/common/ballots/getBallots.ts
```diff
@@ -0,0 +1,53 @@
...
+  const defaultPaginationMetadata = {
+    hasNext: false,
+    totalReturned: 1,
+    nextOffset: 1,
+  };
+
+  const defaultBallots = {
+    ballots: [
+      {
+        address: "0x",
```
stepandel, line 13: `address` in this case is the `submitterAddress`. We should unify these

Flip-Liquid: my read of the draft routes looked like these ballots will live on chain. if that's true, IMO we should include that address as part of the returned content when listing ballots. 

but if the address in the route itself is the submitter's address then i agree

Flip-Liquid:
did a pass after correcting that naming @stepandel, plz take another look when you have a chance

stepandel (approved):
this works!

PR #246, opened 2024-04-22
Michael/agora 1686 copy addresses when clicking on 0x5f4 shorten addresses UI

mcgingras:
resolves AGORA-1686.

I implemented two different approaches, as shown in the [loom](https://www.loom.com/share/464af85696a14fe39000b4fba84a06cf). I'm more of a fan of the second approach, since it offers better visual affordance that the address is copyable, but the first approach is more consistent with the behavior of clicking the discord icon to copy it.

I'm happy to go with either, just let me know!

---

## PR-Codex overview
The focus of this PR is to enhance user experience by adding a copyable feature to addresses and improving click interactions in various components.

### Detailed summary
- Added `copyable` feature to addresses in `DelegateProfileImage`
- Updated click interactions in `VStack` and `HStack` components
- Introduced `CopyableHumanAddress` component for address copying with visual feedback


src/components/shared/CopyableHumanAddress.tsx
stepandel, line 1: ❤️

mcgingras:
I think I'm getting prettier errors which is causing the failed builds... I'll figure out how to get it such that my code is formatting correctly on save. I have a new computer so it's currently not set up 100% yet.

src/components/Layout/Stack.tsx
```diff
@@ -21,6 +21,7 @@ type Props = {
     | "items-start"
     | "items-end"
     | "items-baseline";
+  onClick?: (e?: any) => void;
```
mcgingras, line 24: I found myself wanting to add onClick to an `HStack` and was about to wrap it in a div so I could add an onClick to the div wrapping it. But then I figured the `HStack` is just a div itself... theres no reason it shouldn't just accept an onClick prop. So I'm adding it here... if we don't like this I can take it out!

yitongzhang: lol i've been here a thousand times

yitongzhang:
CleanShot 2024-04-22 at 16 45 00@2x
my only feedback is that this icon is a little thin! does hero icons have a thicker version? if not feel free to grab a thicker one as an svg here: https://www.figma.com/file/KgDS03VEYPclv3l7lJ3Sfm/Agora-Icons?type=design&node-id=181%3A128951&mode=design&t=lnzGu9Gg45a2PNSP-1

Also i know our icon system is a bit of a mess! file it under the list of design system to dos

andreitr (approved):
Before I looked at the code, I couldn't figure out that the address was copyable :) My vote would be to make the icon visible. As for the code—solid work! Congrats on your first commit, Michael

mcgingras:
> CleanShot 2024-04-22 at 16 45 00@2x my only feedback is that this icon is a little thin! does hero icons have a thicker version? if not feel free to grab a thicker one as an svg here: https://www.figma.com/file/KgDS03VEYPclv3l7lJ3Sfm/Agora-Icons?type=design&node-id=181%3A128951&mode=design&t=lnzGu9Gg45a2PNSP-1
> 
> Also i know our icon system is a bit of a mess! file it under the list of design system to dos

I don't think heroicons have a _thicker_ icon, they have a solid, filled icon set but I wasn't a big fan of the solid filled clipboard icon. I can try one of yours though, thanks for the link!

PR #255, opened 2024-04-23
🔐🤏 use bearer token in api client

Flip-Liquid:
## PR-Codex overview
The focus of this PR is to update the AgoraAPI class in `agoraAPI.js` to use a bearer token instead of an API key for authorization.

### Detailed summary
- Updated `constructor` to set `bearerToken` using the API key
- Replaced usages of `apiKey` with `bearerToken` in `GET` and `POST` requests


ferrodri (approved):
LGTM 🚢

PR #265, opened 2024-04-26
[AGORA-1975, AGORA-1973, AGORA-1968, AGORA-1979]🔐🪙 JWT generation, validation

Flip-Liquid:
## PR-Codex overview
This PR updates authentication middleware paths and constants, refactors field names in `AuthInfo`, and adds a new script to generate JWT.

### Detailed summary
- Moved `authenticateApiUser` to `serverAuth` path
- Renamed and updated constants in `constants.ts`
- Refactored field names in `AuthInfo`
- Added `generate-jwt` script for JWT generation

> The following files were skipped due to too many changes: `src/middleware.ts`, `src/app/api/v1/retrofunding/rounds/[roundId]/ballots/[ballotCasterAddressOrEns]/impactMetrics/route.ts`, `src/scripts/generateJwt.ts`, `src/app/lib/auth/serverAuth.ts`, `src/app/lib/auth/edgeAuth.ts`, `yarn.lock`


src/middleware.ts
```diff
@@ -10,15 +10,15 @@ import { hasApiKey } from "@/app/lib/middleware/auth";
   Consider fully validating user api key against postgres pending prisma
   client postgres support on edge runtime.
 */
-export function middleware(request: NextRequest) {
+export async function middleware(request: NextRequest) {
   const path = request.nextUrl.pathname;
   if (path.startsWith("/api/v1") && !path.startsWith("/api/v1/spec")) {
-    const authResponse = hasApiKey(request);
+    const authResponse = await validateBearerToken(request);
```
stepandel, line 16: do we need to validate the token on every request? or this is API key verification?

src/app/lib/auth/edgeAuth.ts
stepandel, line 1: I'm generally not a big fan of combining authentication token & api key verification into the same flow. How would we reconcile the endpoints that require one but not the other?

stepandel (commented):
I'm not a big fan of combining authentication token and API key into the same logic 
How would the r

src/middleware.ts
Flip-Liquid: we will need to validate the token on every request, since it can potentially expire at any point. however, keep in mind this is stateless (i.e. we can validate the token on the edge without accessing our db). 

did you have something else in mind for validating the token?

Flip-Liquid:
> It might be better to split the verification logic into 2 flows (maybe more in the future) depending on the endpoint being called.
> 
> Maybe I'm missing something here but it's difficult to see how we'd manage different permissions this way

Can you explain a little more about what you mean by splitting them into two separate flows? When I look at how authentication is performed now, I do see two separate flows in the sense that we do not issue a JWT to API Key users. I'm up for refactoring to enforce more modularity at the code level, if that's your concern.

whatever validation we do for an arbitrary request in the edge middleware needs to be unified simply by virtue of how the edge runtime works (i.e. it's a single exported function from middleware.ts)

for managing different permissions, I don't think I see the issue you're mentioning, since the scope/permissions for a given user will either be stored in the JWT itself, or associated with their API Key. The flows for resolving the permissions for the supplied JWT or API Key are currently still separate. But, once we have the permissions, we can authorize access to requested resources as appropriate.

for instance, here's how I see it working right now, where the only routes we need to permission are related to ballots:

0) user validates identity with SIWE
1) user is issued JWT containing permissions for the address and chain authenticated with SIWE
2) user supplies JWT to access /ballot endpoints
3) authentication layer validates that the JWT is valid, signed, not expired, and the requested route (i.e. /ballots/{address}) matches the permissions in the supplied token 

For now, API Key users are only able to access "default public" information like delegate statements, votes, proposals, basically the things that anyone can see just be using our frontend.  In the future, as our routes become more complex, we can associate more permissions with a given API key by storing them alongside the user record. 

1) user issued API Key with particular permissions
2) user supplies API Key to access /ballot endpoints
3) authentication layer on the server retrieves permissions associated with API Key, and checks authorization based on the requested resource and the permissions associated with the API Key

N.B. that I haven't yet disallowed access to /ballots for API Key users, so it might become more clear if that's included in this PR too. LMK what you think @stepandel

stepandel:
this is a great explanation, thank you. The part that I was a bit confused about is how you're planing on managing different permissions. I can see the `validateScopeAgainstRoute` function now and encoding scope into JWT is a great approach

PR #268, opened 2024-05-01
[AGORA-1967, AGORA-1962, AGORA-1963] Issue JWT in response to SIWE flow

Flip-Liquid:
## PR-Codex overview
The focus of this PR is to enhance authentication and verification processes in the application.

### Detailed summary
- Added nonce verification in `oas_v1.yaml`
- Updated `generate-typechain` script in `package.json`
- Added new imports and functions in `route.tsx`
- Updated JWT generation and scope logic in `serverAuth.ts`


src/app/api/v1/auth/verify/route.tsx
```diff
@@ -1,11 +1,68 @@
...
+} from "@/app/lib/auth/serverAuth";
 
+// This should create a user id record if there is not one already
+// This should not generate an API key
 export async function POST(request: Request) {
+  const body = await request.json();
+  const { message, signature, nonce } = body;
+
+  // N.B. the below assumes the address we're validating is an
+  // EOA. For EIP-1271, a provider will be necessary
```
stepandel, line 16: we can use https://github.com/voteagora/agora-next/blob/main/src/lib/serverVerifyMessage.ts#L6 for this

src/app/api/v1/auth/verify/route.tsx
```diff
@@ -1,11 +1,68 @@
...
+        chain: {
+          connect: {
+            id: "10",
+          },
+        },
+        enabled: true,
+        description: "Created by SIWE verification",
+      },
+    });
+  }
```
stepandel, line 53: is this just for record keeping? if so, we could do this db call async

src/app/api/v1/auth/verify/route.tsx
```diff
@@ -1,11 +1,68 @@
...
+
+  // create JWT
+  // TODO: resovle scope based on wallet address or user record
+  // TODO: resolve ttl based on wallet address or user record
+  const scope = await getScopeForUser(user.id);
+  const ttl = await getExpiryForUser(user.id);
```
stepandel, line 59: are we expecting this to be different for every user? just want to make sure we're not overthinking this

src/app/api/v1/auth/verify/route.tsx
Flip-Liquid: More or less; the idea is that we create a user ID for monitoring purposes. I can make this more concurrent, since it's not essential for authentication

src/app/api/v1/auth/verify/route.tsx
Flip-Liquid: No, probably not for SIWE routes, this is probably YAGNI for /verify

src/app/api/v1/auth/verify/route.tsx
Flip-Liquid: actually; we do need this in the auth flow to generate the user ID for the JWT request.

src/app/api/v1/auth/verify/route.tsx
stepandel: multi-sigs are gonna be a pretty common use case here

Flip-Liquid:
Bet, I'll send an iteration in that case

src/app/api/v1/auth/verify/route.tsx
Flip-Liquid: addressed this one in https://github.com/voteagora/agora-next/pull/269 by accident @stepandel

stepandel (approved):

PR #269, opened 2024-05-02
[AGORA-1961]🔐 SIWE client integration

Flip-Liquid:
## PR-Codex overview
The focus of this PR is to enhance SIWE (Secure Identity Wallet Environment) functionality by adding SIWE message and signature verification in authentication routes.

### Detailed summary
- Added SIWE message and signature summary in authentication routes
- Extended `serverAuth.ts` with SIWE data for JWT generation
- Updated middleware to validate bearer token for all API routes except excluded ones
- Integrated SIWEProvider in `Web3Provider.tsx` for SIWE functionality
- Implemented SIWE message verification in `auth/verify` route
- Updated SIWE configuration in `SiweProviderConfig.tsx` for SIWE functionality


src/components/shared/SiweProviderConfig.tsx
```diff
@@ -0,0 +1,75 @@
...
+      }),
+    }).then(async (res) => {
+      // save JWT from verify to local storage
+      const token = await res.json();
+      localStorage.setItem(LOCAL_STORAGE_JWT_KEY, JSON.stringify(token));
+      return res.ok;
+    }),
+  getSession: async () => {
+    // return JWT from local storage
+    const session = localStorage.getItem(LOCAL_STORAGE_JWT_KEY);
```
Flip-Liquid, line 56: @stepandel fyi on this piece. the issued JWT maintains the user's session information, so currently this is stateless on the backend/server-side. Retrieving the session simply retrieves the JWT from local storage, and signing out removes it.

Flip-Liquid: let me know if you have strong opinions about this. the [connectkit documentation for SIWE](https://docs.family.co/connectkit/auth-with-custom-backend#siwe-custom-implementation-section-2-configure) is a little strange/misleading. The sample implementation of getSession seems to expect a `/session` endpoint:

```
const siweConfig: SIWEConfig = {
  ...
  getSession: async () => fetch('/api/siwe/session').then((res) => res.ok ? res.json() : null),
  ...
};
```

but does not provide any guidance about how the backend would resolve the session, given that nothing is provided as a parameter to that callback or route (neither in query nor path). Given that, I don't see how one would be able to resolve the session for the current wallet/user _without_ using client storage.

the documentation is also incorrect in that `onSignIn` and `onSignOut` are not members on SIWEConfig FWIW (see [here](https://github.com/family/connectkit/blob/release/v1.7.3/packages/connectkit/src/siwe/SIWEContext.tsx)). given that the documentation is a little wonky, I'm betting that they just didn't consider stateless session mgmt when writing this, so I think the approach in this PR should be okay for now.

stepandel: re: session endpoint. Even though JWT token contains all the data, it can only be decrypted on the server side. So we need this endpoint to decrypt and serve the data back to client.

Since it's gonna be used externally, only our server will be able to decrypt the token

Agree with you on the `onSignOut` `onSignIn` hooks. They're not necessary

Flip-Liquid: Encryption on the JWT is optional (see JWS vs JWE). So whether or not we encrypt is orthogonal. 

I'm amenable to encryption, but if we don't store anything private in the JWT then I'm also ok to keeping it unencrypted, if it saves us implementing /session

Either way, given the signature of the callback, we'll need to keep something in local storage to submit to the /session endpoint.

stepandel: oh interesting learning something new here! I'm okay with keeping it unencrypted in this case  as well then. This simplifies things -- nice!

stepandel (approved):

stepandel (approved):
nice!

PR #277, opened 2024-05-07
don't source control generated contract files

stepandel:


stepandel:
saw you updating `generate-typechain` script in your PR @Flip-Liquid. I'm removing it from source control completely -- these files shouldn't be here in the first place

Flip-Liquid:
nice! thanks for removing these

Flip-Liquid (approved):
LGTM!!!

PR #323, opened 2024-06-07
Fazle/uniswap community page [WIP]

sbvhev:
Community info page by Fazle and Andrei. Most of the functionality should be multi-tenant but there are a few UI configs that are not. 

image


src/app/info/components/ChartGovernanceActiveDelegates.tsx
```diff
@@ -0,0 +1,192 @@
...
+}
+
+type ChartData = {
+  active: string;
+  large: string;
+  date: string;
+};
+
+const ChartGovernanceActiveDelegates = ({ getData }: Props) => {
+  const { primary } = useTenantColorScheme();
```
mcgingras, line 31: I'm doing some work to add custom colors per tenant so we will be able to use that down the line

src/app/info/components/GovernanceCharts.tsx
```diff
@@ -0,0 +1,110 @@
...
+const GovernanceCharts = ({
+  getMetrics,
+  getVotes,
+  getDelegates,
+}: GovernanceChartsProps) => {
+  const tabs = [
+    {
+      value: "top-delegates",
+      title: "Top Delegates",
+      description: "",
```
mcgingras, line 42: does it require a description? Just feels weird to have this empty

src/app/info/components/ChartGovernanceActiveDelegates.tsx
```diff
@@ -0,0 +1,192 @@
...
+import useTenantColorScheme from "@/hooks/useTenantColorScheme";
+import type { MetricTimeSeriesValue } from "@/lib/types";
+import { FREQUENCY_FILTERS } from "@/lib/constants";
+
+interface Props {
+  getData: (
+    metric: string,
+    frequency: string
+  ) => Promise<{ result: MetricTimeSeriesValue[] }>;
+}
```
mcgingras, line 22: is there a reason for passing the `getData` function as a prop vs just defining it here? I already feel like the data fetching function is tightly coupled to this component anyways since it requires metric and frequency. So my first thought if passing the data fetcher down so the chart is abstracted doesn't really stick as much here.

More of a nit / question though, not a deal breaker.

src/app/info/components/GovernanceCharts.tsx
andreitr: @mcgingras ah yes, we used to have description placeholder in the UI but there was no easy way to get them from the back end. I am going to remove them for now.

src/app/info/components/ChartGovernanceActiveDelegates.tsx
andreitr: @mcgingras getData is a server function that get's passed to the client component. I guess that's the only reason. I agree that having the data method inside the chart would be a better design.

andreitr (approved):

src/app/info/components/ChartGovernanceActiveDelegates.tsx
mcgingras: ah okay, thanks!

PR #332, opened 2024-06-11
Scroll token claim flow

mcgingras:
tbd wip
opening early for visibility


jefag:
I didn't see this PR until after I did #459.

Instead of reconciling, I think it makes sense to get the prisma working in isolation, merge to main, and then you remove that from your PR.

jefag:
...reason being is we want the scroll schema into the repo, before the token claim flow is merged.

package-lock.json
stepandel, line 1: pls don't commit `package-lock.json` ty!

package.json
```diff
@@ -57,8 +59,11 @@
     "class-variance-authority": "^0.6.1",
     "clsx": "^1.2.1",
     "connectkit": "^1.5.3",
+    "cookies-next": "^4.2.1",
     "csstype": "^3.0.10",
+    "csv-writer": "^1.6.0",
```
mcgingras, line 64: can remove some of these when we merge, these are just for the script, which Jeff is going to rewrite

PR #338, opened 2024-06-12
dirty fix for getting the right number of delegators

stepandel:
## PR-Codex overview
This PR focuses on optimizing the `getDelegates` function by using a cached amount when the recalculation is expensive, but only if the namespace is "optimism".

### Detailed summary
- Added a condition to use cached amount only if namespace is "optimism"


stepandel:
A clean fix incoming.
We'll need to add extra views per tenant in order to have a better tenant strategy

andreitr (approved):

PR #342, opened 2024-06-13
Adds tenant theming

mcgingras:
__200 files changed!
 Wow!__

### Description
This branch removes all module based css and converges all color based css into semantic tailwind names. 

### How customization works

1. Add custom font names to tailwind config where the value is a css variable
`agoraPrimary: "var(--agora-primary)"`

2. add custom colors to the tenant
```
customization: {
    primaryColor: "aqua",
 }
 ```
 
3. get the color from the tenant
`const primaryColor = ui?.customization?.primaryColor || defaults.primaryColor;`

4. inject that color into the css var that tailwind is set to look for
`const style = { "--agora-primary": primaryColor } as React.CSSProperties;`

5. set that style at the root of the app so the variable is set throughout the app
`style={style}`

So we are defining tailwind classes that are looking for css variables. Then we pull styles from the tenant and set the css variables to those styles at the root of the app. Then the tailwind styles are attached to the custom styles defined by those variables. This way we can use a single tailwind className `text-primaryColor` and be able to swap out what that color actually is on a per tenant basis.

### CODE REVIEW

File: src/lib/tenant/tenantUI.ts [hunk 1]
```diff
@@ -65,6 +65,19 @@ type TenantUIParams = {
...
+    tertiary?: string;
+    neutral?: string;
+    wash?: string;
+    line?: string;
+    positive?: string;
+    negative?: string;
+    brandPrimary?: string;
+    brandSecondary?: string;
+    font?: string;
+  };
```
mcgingras, line 80: @andreitr would be good to get your feedback here if you think this theming setup thought tenants makes sense

src/app/layout.tsx
```diff
@@ -19,15 +21,58 @@ async function fetchDaoMetrics() {
...
+    "--tertiary": tertiary,
+    "--neutral": neutral,
+    "--wash": wash,
+    "--line": line,
+    "--positive": positive,
+    "--negative": negative,
+    "--brand-primary": brandPrimary,
+    "--brand-secondary": brandSecondary,
+    "font-family": font,
+  } as React.CSSProperties;
```
mcgingras, line 72: @andreitr here is the other place I'd like to get your feedback on -- this is the other half of how we pull the theme from the tenant. 

If you agree with these two files, the rest of the code is just ripping out css and updating them to use these theme variables. I might not have caught 100% of them but I think I did a decent job and we can always catch the rest as we see them later.

Goal is for no glaringly obvious style degradations

src/app/layout.tsx
```diff
@@ -19,15 +21,58 @@ async function fetchDaoMetrics() {
...
+  wash: "#FAFAFA",
+  line: "#EAEAEA",
+  positive: "#00992B",
+  negative: "#C52F00",
+  brandPrimary: "#171717",
+  brandSecondary: "#FFFFFF",
+};
+
+const defaults = standard;
+
```
andreitr, line 38: What's the reason for reassigning `const defaults = standard`?

src/app/layout.tsx
andreitr: I think this makes sense. My initial reaction was to return `React.CSSProperties` from the tenant but I see why you chose to assign these values in the layout. LGTM

src/components/Button.jsx
```diff
@@ -0,0 +1,25 @@
+import Link from "next/link";
```
andreitr, line 1: Nice!

src/components/Proposals/ProposalPage/CastVoteDialog/CastVoteDialog.tsx
```diff
@@ -217,12 +234,15 @@ export function SuccessMessage({
       <div className="mb-2 text-2xl font-black">
         Your vote has been submitted!
       </div>
-      <div className="mb-5 text-sm text-gray-700">
+      <div className="mb-5 text-sm text-secondary">
         It might take up to a minute for the changes to be reflected. Thank you
         for participating in Optimism’s token house.
```
andreitr, line 239: Another optimism reference.

src/components/Delegates/DelegateVotes/DelegateVotes.jsx
```diff
@@ -100,14 +99,14 @@ export default function DelegateVotes({ fetchDelegateVotes }) {
...
-              <div className={styles.details_container}>
-                <VStack className={styles.details_sub}>
+              <div>
+                <VStack className="py-4 px-6">
                   <HStack justifyContent="justify-between" gap={2}>
                     <VStack>
-                      <span className="text-[#66676b] text-xs font-medium">
+                      <span className="text-tertiary text-xs font-medium">
                         {`${propHeader(vote)} with ${pluralizeVote(vote.weight, "optimism")}`}
                       </span>
```
andreitr, line 108: Note to self - nuke all optimism references.

src/lib/tenant/tenantUI.ts
andreitr: I think this makes sense. It would be good to give this template to Josiah so he can do this along with tenant assets.

src/components/Events/EventFeed.jsx
```diff
@@ -0,0 +1,116 @@
+"use client";
```
andreitr, line 1: Is this used?

andreitr (approved):

PR #369, opened 2024-06-27
Add latest to set of options available for `/balances` and `/analytics/metric`

jefag:
The approach here, is basically just crop what gets sent in the 24h request.  We can speed these up after, without involving front-end.  

---

## PR-Codex overview
The focus of this PR is to update the `getTreasuryBalanceTS` and `getMetricTS` functions to return a promise with a specific result object when the frequency is "latest".

### Detailed summary
- Updated `getTreasuryBalanceTS` to return a promise with `{ result: TokenBalance[] }` when frequency is "latest"
- Updated `getMetricTS` to return a promise with `{ result: MetricValue[] }` when frequency is "latest"
- Refactored return statements to include the specific data type for clarity

### DIFF
--- a/src/app/api/analytics/metric/[metric_id]/[frequency]/route.ts
+++ b/src/app/api/analytics/metric/[metric_id]/[frequency]/route.ts
@@ -12,8 +12,17 @@ type MetricValue = {
   value: any;
 };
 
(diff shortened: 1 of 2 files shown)


andreitr (approved):
LGTM

PR #377, opened 2024-06-28
Rename `votes` -> `voter_count`

jefag:
...just trying to avoid massive confusion.

Happy to merge this after we merge the #323 , and presumably patch the usage of `votes` key in the same merge.

I need a manual DB migration to go simultaneously.


andreitr (approved):
LGTM

PR #380, opened 2024-07-01
[Uniswap] Fix delegates queries

stepandel:


src/lib/tenant/tenantSlugFactory.ts
```diff
@@ -13,7 +13,7 @@ export default class TenantSlugFactory {
         return "ETHERFI";
 
       case TENANT_NAMESPACES.UNISWAP:
-        return "UNISWAP";
+        return "UNI";
```
stepandel, line 16: @andreitr fyi I updated slug to UNI, please check if this breaks anything on FE. This is how it's defined on the DB

src/app/lib/provider.ts
```diff
@@ -41,6 +44,7 @@ if (isProd) {
     global.ethProvider = MulticallWrapper.wrap(
       getDefaultProvider("mainnet", {
         alchemy: alchemyId,
+        exclusive: "alchemy",
```
stepandel, line 47: @andreitr also fyi, I added this to mute ethers Infura errors -- this should work fine so long as we don't have a fallback provider

andreitr: @stepandel sweet. I fixed this issue in the Scroll branch by going to tenant-specific Alchemy or JsonRPC providers but it will be good to have this in place in the meantime.

andreitr (approved):
LGTM

PR #385, opened 2024-07-05
Andrei/agora 2209 add support for stakeholders i represent in delegate

andreitr:
- Stakeholders are defined in the tenant file 
- Stakeholders displayed on the delegate page 
- Stakeholders can be added
- Backwards compatible with existing Uniswap delegate statements

Screenshot 2024-07-05 at 1 03 06 PM
Screenshot 2024-07-05 at 1 02 15 PM


src/components/DelegateStatement/CurrentDelegateStatement.tsx
```diff
@@ -140,8 +170,8 @@ export default function CurrentDelegateStatement() {
     }
   }, [address, reset]);
 
-  if (!isConnected) {
-    return <ResourceNotFound message="Oops! Nothing's here" />;
+  if (!isConnected && !isConnecting) {
+    router.push("/");
   }
```
mcgingras, line 175: this pushes you back home if you are not connected? I feel like this might be jarring if you are not expecting it.

src/components/DelegateStatement/CurrentDelegateStatement.tsx
```diff
@@ -103,6 +113,26 @@ export default function CurrentDelegateStatement() {
...
             )?.topIssues
           : defaultIssues,
+
+      topStakeholders:
+        (
+          delegateStatement?.payload as {
+            topStakeholders: {
+              value: string;
+              type: string;
+            }[];
```
mcgingras, line 123: is there anyway to type this better from wherever it is fetched? Having to add all of these `as` types makes it a bit tougher to read. Not a huge deal though, just a nit

src/components/DelegateStatement/DelegateStatementForm.tsx
```diff
@@ -64,7 +62,7 @@ export default function DelegateStatementForm({
     if (!walletClient) {
       throw new Error("signer not available");
     }
-    values.topIssues = cleanTopIssues(values.topIssues);
+    values.topIssues = values.topIssues.filter((issue) => issue.value !== "");
```
mcgingras, line 65: similar to feedback you gave me before about mutating the object -- I don't think it would effect behavior but could probably do something like

```
const updatedValues = {
...values,
topIssues: values.topIssues.filter((issue) => issue.value !== "")
}
```

then use updatedValues in place of values

src/components/DelegateStatement/TopIssuesFormSection.tsx
```diff
@@ -33,22 +28,22 @@ export default function TopIssuesFormSection({
   form,
 }: TopIssuesFormSectionProps) {
   const { ui } = Tenant.current();
-  const topIssues = useWatch({ name: "topIssues" });
+  const topIssues = useWatch({ name: FORM_KEY });
```
mcgingras, line 31: I like the use of const instead of a string but `FORM_KEY` feels like it could potentially be a confusing variable name especially if we had a component with many useWatches

mcgingras (approved):
👍

PR #386, opened 2024-07-05
hotfix: proposal/create is not supported for namespace error

stepandel:


andreitr (approved):
I was about to fix that!!! thanks for jumping in. Btw, auto deploys are disabled for Uniswap FYI.

PR #400, opened 2024-07-16
Add snapshot votes ens

jefag:


src/app/api/common/votes/getVotes.ts
```diff
@@ -18,6 +18,55 @@ const getVotesForDelegate = ({
...
+  address,
+  page = 1,
+}: {
+  address: string;
+  page: number;
+}) {
+  const { slug } = Tenant.current();
+  const pageSize = 10;
+
+  const { meta, data: votes } = await paginateResult(
```
mcgingras, line 31: If we have prisma is there a reason for doing queryRaw here vs just using prisma like an ORM as intended with something like

```
prisma.snapshotVotes.findMany({
  where: {
    dao_slug: slug
  },
  select: {
     ....you get the point
  }
})
```

any perf hit for this or whats the reason for queryRaw? 
personal pref but I find the prisma version easier to read (I'm used to it)
and I don't see why we are using prisma at all if we are just going to subvert it all of the time with queryRaw, right?

jefag: No, prisma is only there if you want it for some to-be-written access function.

My personal pref is the raw.  And also, my prediction is that we're going to outgrow it completely. Just a question of when.

jefag:
Don't look at this PR like something finished; look at it as boiler plate that needs to be adjusted for specific features, in order to reach feature parity.  Kindof, passing you the ball here.

src/lib/tokenUtils.ts
```diff
@@ -34,6 +34,17 @@ export function pluralizeVote(count: BigInt) {
     .join("")} votes`;
 }
 
+export function pluralizeSnapshotVote(count: bigint) {
```
andreitr, line 37: Can this be abstracted to just pluralizeVote?

andreitr (approved):
LGTM! Left one non-blocking comment.

src/lib/tokenUtils.ts
mcgingras: The problem is we already have a `pluralizeVote` in this file, but it's expecting a different unit so it was breaking for snapshot votes. Hence the snapshot specific votes. I agree it would be nice to have a single function but would probably require a change more upstream where the data is indexing.

PR #414, opened 2024-07-23
Andrei/agora 2297 cyber tenant

andreitr:
Working Cyber tenant.

Screenshot 2024-07-23 at 3 35 40 PM


src/app/delegates/components/EndorsedFilter.tsx
```diff
@@ -42,14 +42,14 @@ export default function EndorsedFilter() {
   return (
     <Listbox
       as="div"
-      value={filterParam || endorsedFilterOptions.all.value}
+      value={filterParam || endorsedFilterOptions.false.value}
       onChange={(value) => handleChange(value)}
     >
       {() => (
         <>
           <Listbox.Button className="w-full sm:w-fit bg-[#F7F7F7] text-base font-medium border-none rounded-full py-2 px-4 flex items-center">
```
mcgingras, line 50: now that the style branch is merged can we change `bg-[#F7F7F7]` to `bg-wash`

src/app/delegates/components/EndorsedFilter.tsx
```diff
@@ -42,14 +42,14 @@ export default function EndorsedFilter() {
...
             {endorsedFilterOptions[filterParam]?.value ||
-              endorsedFilterOptions.all.value}
+              endorsedFilterOptions.true.value}
             <ChevronDown className="h-4 w-4 ml-[2px] opacity-30 hover:opacity-100" />
           </Listbox.Button>
           <Listbox.Options className="mt-3 absolute bg-[#F7F7F7] border border-[#ebebeb] p-2 rounded-2xl flex flex-col gap-1 z-20 w-max">
```
mcgingras, line 55: same as above, but also `border-[#ebebeb]` to `border-line`

src/app/lib/ENSUtils.ts
```diff
@@ -1,16 +1,19 @@
-import { ethProvider } from "@/app/lib/provider";
 import { truncateAddress } from "@/app/lib/utils/text";
 import { isAddress } from "viem";
 import { cache } from "react";
+import { AlchemyProvider } from "ethers";
```
mcgingras, line 4: is there a reason we don't just use the wagmi hooks for ens

src/assets/tenant/cyber_delegate.svg
mcgingras, line 1: cyber_delegate_placeholder

one of the mocks I saw had this -- but I don't think it matters too much

src/lib/tenant/configs/contracts/ens.ts
```diff
@@ -5,11 +5,19 @@ import {
...
 
-export const ensTenantContractConfig = (isProd: boolean): TenantContracts => {
+interface Props {
+  isProd: boolean;
+  alchemyId: string;
+}
+
+export const ensTenantContractConfig = ({
+  isProd,
+  alchemyId,
```
mcgingras, line 19: do we think our alchemyId will change between tenants or can we just do something like `process.env.ALCHEMY_ID` rather than passing it through to each tenant's config?

mcgingras (approved):

src/lib/tenant/configs/contracts/ens.ts
andreitr: I don't think so. I also don't have a strong preference here but my initial approach was to minimize the number of places in which `process.env.` vars are referenced.

src/assets/tenant/cyber_delegate.svg
andreitr: Pretty sure the latest decision was to use the delegate avatar. I can check.

src/app/lib/ENSUtils.ts
andreitr: Good point. No idea!

PR #420, opened 2024-07-24
Add Delegate Satement Creators w/ zero  VP to delegates page

jefag:


jefag:
@stepandel, @andreitr , this patch [db7458e](https://github.com/voteagora/agora-next/pull/420/commits/db7458ed19bbec842f202625df34f428aa0b5036) is the heart of it.  everything else is re-factoring so that I could iterate faster.

andreitr (commented):
Looks good. I am going to do more QA tomorrow but at the first glance LGTM. I also noticed duplicate results but that's probably due to the fact that you haven't merged the latest.

src/app/api/common/delegates/getDelegates.ts
```diff
@@ -95,149 +95,157 @@ async function getDelegates({
       )`
       : "";
 
+  const includeZeroVPDelegates = true;
+
+  let delegateUniverseCTE: string;
+
+  const tokenAddress = contracts.token.address;
+
+  if (includeZeroVPDelegates) {
```
stepandel, line 104: do we need this condition? seems like it's hardcoded to `true`. Also I don't really see a use case where we wouldn't include it

stepandel (approved):
I like this approach -- functionality looks good. just a couple of points to keep the code clean
could you also remove console.log s please

src/app/api/common/delegates/getDelegates.ts
```diff
@@ -95,149 +95,157 @@ async function getDelegates({
       )`
       : "";
 
+  const includeZeroVPDelegates = true;
+
+  let delegateUniverseCTE: string;
+
+  const tokenAddress = contracts.token.address;
```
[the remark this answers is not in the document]
andreitr: @stepandel, TenantContract  class converts the address to lower case in the getter 

```
  get address(): string {
    return this._address.toLowerCase();
  }
```

src/app/api/common/delegates/getDelegates.ts
jefag: It was only there as a backup.  to show the logic to undo this, if performance was terrible for one tenant.

jefag: I'll back it out in the rebase.  This thread is going to get hidden when I do so.

PR #421, opened 2024-07-25
Resolves citizen issue

mcgingras:


jefag (approved):

jefag:
Oh, wait, should check on vercel that it actually fixes it before merging.

jefag (approved):
Tried Optimism just now.  Definitely fixes the issue.

PR #426, opened 2024-07-26
Andrei/agora 2302 cyber og attributes

andreitr:
Key changes
- Generic OG Preview image background
- Reverted stylistic changes for OG Images (cc @mcgingras) 
- Added code of conduct for Cyber 
- Code of conduct link pulled from the tenant config 
- Delegate statement preview link is pulled from the tenant config
- Misc code cleanup

image


src/app/api/images/og/assets/shared.tsx
```diff
@@ -7,7 +7,7 @@ export const LogoPill = () => {
 
   return (
     <div tw="flex">
-      <div tw="flex flex-row items-center border border-line rounded-full px-[26px] py-[14px]">
```
andreitr, line 10: @mcgingras I reverted this change because the styles were not showing up in the generated image. We can debug this when we need to offer more customizations.

mcgingras: No problem! This is pretty specific part of the app so I think it's okay if it's default themed

mcgingras (approved):

PR #427, opened 2024-07-29
Social links

andreitr:
- Resized social icons
- Updated Twitter icon to X

Screenshot 2024-07-29 at 11 05 35 AM


mcgingras (approved):
RIP twitter, I'll never call it x

src/components/Delegates/DelegateCard/DelegateActions.tsx
```diff
@@ -46,9 +45,8 @@ export function DelegateActions({
   }
 
   return (
-    <HStack
-      alignItems="items-stretch"
-      className={className ? className + "justify-between" : "justify-between"}
+    <div
+      className={`flex flex-row items-stretch ${className ? className + "justify-between" : "justify-between"}`}
```
mcgingras, line 49: could probably add `justifyBetween` outside of the ternary since it's used by both

andreitr: @mcgingras good catch. Will fix.

PR #429, opened 2024-07-29
Tenant-specific Google Analytics

andreitr:
- Added GA 
- Nuked tenant color attribute


mcgingras (commented):
do we need the color? I get that we can pull it from customization now, but looks like those were not added to tenants where you removed color -- just not sure if we need that as a single var or not anymore

mcgingras (approved):

PR #450, opened 2024-08-14
Set quorum to 30% for Cyber

jefag:


jefag:
Is this the right implementation of the spirit of this tactical solution?

jefag:
And if so, is it expected that `getProposalStatus` thinks everything is defeated?  Friendly reminder, total voteable supply is calculated from the main deployment delegations (not test deployment).

image

andreitr (approved):

jefag:
@andreitr does the all-defeated make sense to you?

PR #461, opened 2024-08-19
Andrei/agora 2416 as a proposer i must be able to queue and execute a proposal

andreitr:
- `queued_block` column added to track when a prop was queued
- `proposal-execute` flag for the tenant 
- added timelock to contracts 

Screenshot 2024-08-26 at 9 29 19 AM


mcgingras (approved):
looks good! I like the clean separation between the buttons

PR #463, opened 2024-08-21
Scroll Tenant

andreitr:
Screenshot 2024-08-21 at 3 13 09 PM


src/app/api/common/metrics/getMetrics.ts
```diff
@@ -5,12 +5,22 @@ import { cache } from "react";
...
+  try {
+    const votableSupply = await prisma[`${namespace}VotableSupply`].findFirst(
+      {}
+    );
+    return {
+      votableSupply: votableSupply?.votable_supply || "0",
+      totalSupply: totalSupply.toString(),
+    };
+  } catch (e) {
+    // Handle prisma errors for new tenants
```
andreitr, line 18: This handles the new tenant edge case. Basically return a zero before votable supply is indexed.

src/app/api/common/votableSupply/getVotableSupply.ts
```diff
@@ -4,11 +4,18 @@ import { cache } from "react";
...
+    const votableSupply = await prisma[`${namespace}VotableSupply`].findFirst(
+      {}
+    );
+    return votableSupply?.votable_supply || "0";
+  } catch (error) {
+    console.log("Error fetching votable supply:", error);
+
+    // Handle prisma errors for new tenants when votable supply is not available
+    // 1 is returned to avoid zero division errors
+    return "1";
```
andreitr, line 17: Same here. New tenant edge case. Need to return 1 in order to avoid zero division error.

src/app/proposals/draft/[id]/page.tsx
```diff
@@ -28,6 +29,7 @@ export default async function DraftProposalPage({
   }
 
   const draftProposal = await fetchDraftProposal(parseInt(params.id));
+  const proposalTypes = await fetchProposalTypes();
```
Comment (andreitr) [inline src/app/proposals/draft/[id]/page.tsx:32]: @mcgingras I was getting a TypeScript error here. It was complaining that `DraftFormServer.tsx` was not returning a valid JSX in the suspense state. Anyway, this is a quick and dirty workaround. I am also now realizing that we don't even need the `DraftFormServer.tsx` wrapper.

Comment (mcgingras) [reply src/app/proposals/draft/[id]/page.tsx:32]: Oh interesting. I was getting that error too sometimes but it would still build, I was chalking it up to my tsconfig possibly not being up to date with server components. Yea, maybe we don't need it, I think the intention was having a wrapper to allow us to fetch proposal types while the rest of the real component could be a client component

mcgingras (approved):

PR #474, opened 2024-08-28
Tries to resolve social proposal creation

mcgingras:


src/app/proposals/draft/components/DraftPreview.tsx
```diff
@@ -35,24 +35,28 @@ const DraftPreview = ({
   const { data: threshold } = useProposalThreshold();
   const { data: manager } = useManager();
   const { data: blockNumber } = useBlockNumber();
-  const { data: accountVotesData } = useGetVotes({
-    address: address as `0x${string}`,
-    blockNumber: blockNumber || BigInt(0),
-  });
```
mcgingras, line 41: weird caching issues with this query

src/app/proposals/draft/utils/createSnapshot.ts
```diff
@@ -66,7 +66,7 @@ export async function createSnapshot({
 }) {
   const tenant = Tenant.current();
   const walletClient = getWalletClient(tenant.contracts.token.chain.id);
-  const publicClient = getPublicClient(tenant.contracts.token.chain.id);
+  const publicClient = getPublicClient(1);
```
mcgingras, line 69: needs to be 1, even on testnet it tries to read current block so we can't use testnet block

src/lib/tenant/configs/contracts/ens.ts
```diff
@@ -36,7 +36,7 @@ export const ensTenantContractConfig = ({
     token: new TenantContract<ITokenContract>({
       abi: ERC20__factory.abi,
       address: TOKEN as `0x${string}`,
-      chain: mainnet,
+      chain: chain,
```
mcgingras, line 39: chain should not be hardcoded to mainnet

PR #475, opened 2024-08-28
IP Blocker page wrapper 

andreitr:
- Checks your origin county and region against the block list 
- Displays a placeholder page

**Usage**
Wrap contents on any top-level page with `<IPBlocker enabled={true}>YOUR Content</IPBlocker>`

**Future Improvements**
To make this multi-tenant, we can wrap every page inside in the IPBlocker and set the enabled flag in the tenant. 

Screenshot 2024-08-28 at 3 16 32 PM


src/components/shared/IPBlocker.tsx
```diff
@@ -0,0 +1,60 @@
...
+import Tenant from "@/lib/tenant/tenant";
+import Image from "next/image";
+import { Button } from "@/components/ui/button";
+
+// Cuba, Iran, North Korea, Syria, and Russia are blocked from participating in the airdrop
+const BLOCKED_COUNTRIES = ["CU", "IR", "KP", "SY", "RU"];
+
+// The following ISO 3166-2 codes are for Crimea, Donetsk and Luhansk regions of occupied Ukraine
+// https://www.iso.org/obp/ui/#iso:code:3166:UA
+const BLOCKED_REGIONS = ["UA-43", "UA-14", "UA-09"];
```
mcgingras, line 12: do you think we would want to make this tenant specific? Probably fine for now since scroll is the only airdrop we are interested in gating now.

src/components/shared/IPBlocker.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+interface Props {
+  children: ReactNode;
+  enabled: boolean;
+}
+
+const ipData = new IPData(process.env.IP_DATA_API_KEY!, {
+  maxAge: 3600,
```
mcgingras, line 20: what does the `maxAge` flag represent here? Would it ever expire? I feel like we'd want IP to be blocked indefinitely right?

mcgingras (approved):
looks good, just wondering about eventually moving to tenant

src/components/shared/IPBlocker.tsx
```diff
@@ -0,0 +1,60 @@
...
+const AccessRestricted = () => {
+  const { ui } = Tenant.current();
+
+  return (
+    <div className="flex flex-col w-full h-screen items-center justify-center gap-2">
+      <Image src={ui.logo} alt="logo" width="68" height="68" />
+
+      <div className="font-semibold text-xl mt-4">Access Restricted</div>
+      <div className="text-sm">
+        Unfortunately, the airdrop is not available in your region.
```
mcgingras, line 53: I think we might want "claim flow" instead of "airdrop" -- cc Yitong about that though.

mcgingras: I think scroll was trying to avoid airdrop

src/components/shared/IPBlocker.tsx
andreitr: @mcgingras the maxAge header determines how often to refetch the up data.

PR #479, opened 2024-09-02
remove rpgf 4 banner and fix favicon

yitongzhang:
## Improve favicon
before: chunky, does not work in dark mode, gets cropped by arc browser

CleanShot 2024-09-02 at 16 12 44@2x

after: very demure, very mindful

CleanShot 2024-09-02 at 16 13 36@2x

by the way, am i crazy or does the agora-next app not explicitly define favicon? @andreitr tagging you to make sure i didn't mess up lol

## Remove banner
CleanShot 2024-09-02 at 16 00 57@2x


public/favicon/safari-pinned-tab.svg
mcgingras, line 1: this just looks like a big black square, is it supposed to be the favicon?

mcgingras (approved):
lgtm lets gooo

PR #481, opened 2024-09-03
Delegation setup flow

andreitr:
Changes to enable Scroll delegation 

- zero addresses for token, governor, treasury 
- remove cards for the info page header
- updated getMetrics to fail gracefully 
- updated useConnectedDelegate to fail gracefully
- disabled token delegation
- updated top issues
- updated top issues icons


mcgingras (approved):
looks good

PR #488, opened 2024-09-06
Cycling through RPC urls

andreitr:
In theory, this should at least distribute the load between the two RPC endpoints.


mcgingras:
lol I was logged in with the agora dev account

mcgingras (approved):

PR #491, opened 2024-09-09
Andrei/agora 2625 add a cta to the airdrop page from the info page

andreitr:
Screenshot 2024-09-09 at 11 34 10 AM


src/app/info/components/InfoHero.tsx
```diff
@@ -19,6 +20,14 @@ export const InfoHero = () => {
         </h1>
         <p className="text-base font-medium text-gray-4f mt-2 sm:mt-0 ">
```
mcgingras, line 21: we could probably command-f for `text-gray-4f` and change that to one of the custom colors

src/app/info/components/InfoHero.tsx
```diff
@@ -40,11 +49,11 @@ export const InfoHero = () => {
...
-  image,
-}: {
+                className,
+                link,
+                linkText,
+                image,
+              }: {
   className?: string;
   link: string;
   linkText: string;
```
mcgingras, line 59: same with line 64 -- I'd change to bg-neutral border-line

mcgingras (approved):
looks good but I notice that there are still some components that are not up to date on the latest custom style classNames

PR #493, opened 2024-09-09
Updated scroll top issues + hovers

andreitr:
Screenshot 2024-09-09 at 3 10 22 PM


src/components/DelegateStatement/TopStakeholdersFormSection.tsx
```diff
@@ -55,7 +55,7 @@ export default function TopStakeholdersFormSection({
         <h3 className="font-bold">Stakeholders I represent</h3>
         <DropdownMenu>
           <DropdownMenuTrigger
-            className={`${canAddMoreStakeholders ? "text-tertiary" : "text-gray-500"} outline-none`}
+            className={`${canAddMoreStakeholders ? "text-tertiary cursor-pointer" : "text-gray-500"} outline-none`}
```
mcgingras, line 58: is `text-gray-500` meant to be a subdued version of `text-tertiary`? If so, we can use something like `text-tertiary/80` or something

mcgingras (approved):

PR #495, opened 2024-09-10
fix: Updated Sentry config to remove monitoring error

kent:
Fix config of nextjs, and Sentry monitor

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,11 +1,8 @@
 /** @type {import('next').NextConfig} */
 
 const path = require("path");
-module.exports = {
-  sassOptions: {
-    includePaths: [path.join(__dirname, "styles")],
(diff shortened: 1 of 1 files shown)


next.config.js
```diff
@@ -1,11 +1,8 @@
 /** @type {import('next').NextConfig} */
 
 const path = require("path");
-module.exports = {
-  sassOptions: {
-    includePaths: [path.join(__dirname, "styles")],
-  },
-};
```
mcgingras, line 8: is this going to break our ability to use styles file? I know we are all in on tailwind, but we still have a bit in styles, like custom fonts and some things that are challenging to style via tailwind

andreitr (approved):
LGTM

PR #496, opened 2024-09-10
Add partial delegation support

jefag:
Screenshot 2024-09-16 at 12 20 01 PM


src/app/api/common/delegates/getDelegates.ts
```diff
@@ -350,34 +351,45 @@ async function getDelegate(addressOrENSName: string): Promise<Delegate> {
...
-          WHERE to_delegate=$1 OR from_delegate=$1
-        ) t;
-        `,
-        address,
-        contracts.alligator?.address
-      )
-    : prisma.$queryRawUnsafe<{ num_of_delegators: BigInt }[]>(
-        `
-        SELECT
+  const numOfAdvancedDelegationsQuery = `SELECT count(*)
```
mcgingras, line 354: man it'd be nice to get rid of alligator 😅

src/components/Delegates/DelegateCard/DelegateActions.tsx
```diff
@@ -36,6 +37,36 @@ export function DelegateActions({
     delegate.address.toLowerCase() as `0x${string}`
   );
 
+  const delegationButton = () => {
+    // TODO: Replace namespace lookup with governor-level switch in the tenant
+    switch (namespace) {
+      case TENANT_NAMESPACES.SCROLL:
```
mcgingras, line 43: It would be great if at some point we could base these switches off of the type of governor, rather than the namespace. Is that possible or does that not make sense given the context here?

mcgingras: (not a thing for this PR, just in general, I think if possible it's better to switch on governor type since its more general than tenant)

src/components/Delegates/DelegateCard/PartialDelegateButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  return (
+    <Button
+      onClick={(e: any) => {
+        e.preventDefault();
+        openDialog({
+          type: "PARTIAL_DELEGATE",
+          params: {
+            delegate,
+            fetchCurrentDelegatees,
```
mcgingras, line 22: man, this is one of my least favorite things about server components. I'm guessing we cannot just import fetchCurrentDelegates from the modal itself, because the modal is a client component and we can't import server actions in a client component, is this correct? So instead we have to import the server action higher up, then pass it down via props. It's a confusing pattern for me. I feel like the best devx or ergonomics would be being able to write the component, and import the server action where you need it. 

Not a comment on your code, maybe there's nothing we can do about it. But if there is a way for us to solve this pattern a bit better, I think it would be awesome. Maybe it's impossible though

src/components/Delegates/Delegations/DelegationsContainer.tsx
```diff
@@ -83,18 +80,10 @@ function DelegationsContainer({
             <Table className="min-w-full">
               <TableHeader className="text-xs text-secondary sticky top-0 bg-white z-10">
```
mcgingras, line 81: maybe we could do `bg-neutral` here, sorry to be such a pest with these color changes

src/lib/tenant/configs/contracts/scroll.ts
```diff
@@ -20,7 +21,8 @@ export const scrollTenantContractConfig = ({
   isProd,
   alchemyId,
 }: Props): TenantContracts => {
-  const TOKEN = ZERO_ADDRESS;
+  const TOKEN = process.env.NEXT_PUBLIC_SCROLL_TOKEN_ADDRESS!;
```
mcgingras, line 24: Is there a reason we need to pass this down via env rather than hardcoding it into the tenant config like we typically do with the other tenants?

src/lib/tenant/configs/contracts/scroll.ts
```diff
@@ -1,5 +1,6 @@
 import {
   AgoraGovernor__factory,
+  AgoraToken__factory,
```
mcgingras, line 3: is `AgoraToken` the same as `L2GovToken`?

src/lib/blockTimes.ts
```diff
@@ -73,7 +73,7 @@ export function getHumanBlockTime(
     case 534352:
       const blockSeconds = getSecondsPerBlock();
       const estScrollSecondsDiff =
-        Number(latestBlock.number) - Number(blockNumber) * blockSeconds;
+        (Number(latestBlock.number) - Number(blockNumber)) * blockSeconds;
```
mcgingras, line 76: nice catch!

src/components/Dialogs/PartialDelegateDialog/PartialDelegationSuccess.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <div className="w-full">
+        <Image
+          className="w-full"
+          src={ui.assets.success}
+          alt="Delegation successful image"
+        />
+      </div>
+
+      <h1 className="mt-4 mb-2 text-2xl font-extrabold text-primary">
+        You&apos;ve delegated your votes!
```
mcgingras, line 24: I always forget &apos; so I've just been wrapping in `` like {`You've delegated your votes`} not sure if you'd like that pattern also!

mcgingras (approved):

src/components/Delegates/DelegateCard/DelegateActions.tsx
jefag: IMHO, we should be doing it now, in this PR.  

There is a couple other spots where it's `namespace == SCROLL` added in this PR, but really it's GOV_TYPE == 'AGORA_V1'

andreitr: @jefag I didn't want to overload this PR with a bunch of tenant work but yes, it makes sense to introduce a governor-level switch.

src/components/Delegates/DelegateCard/PartialDelegateButton.tsx
andreitr: Completely agree. This trips me up every time. Maybe we can start wrapping some of these ss calls into hooks?

src/lib/tenant/configs/contracts/scroll.ts
andreitr: yep, happy to rename it to L2GovToken if it makes more sense.

PR #528, opened 2024-09-30
Fix root path rendering issue in middleware

mcgingras:
# Fix Root Path Rendering Issue in Middleware

## Problem
The root path ("/") was not rendering properly, instead displaying the entire HTML content wrapped in a `<pre>` tag. This was caused by the middleware applying CORS headers and setting the `content-type` to `application/json` for all routes, including non-API routes.

## Solution
Modified the middleware to only apply CORS headers to API routes (those starting with the `API_PREFIX`). For non-API routes, including the root path, the middleware now simply calls `NextResponse.next()` without modifying the headers.

## Changes
- Updated the `middleware.ts` file to conditionally apply CORS headers.
- Ensured that the `content-type` header is not set to `application/json` for non-API routes.


stepandel (approved):
Sorry! thank you for fixing!!

PR #530, opened 2024-09-30
Queue, Cancel, and Executure proposals on OZ, Bravo and Agora Governors

andreitr:
Functionality per governor 

Agora (Scroll, Cyber)
- Cancel 
- Queue 
- Execture

OZ (ENS)
- Queue 
- Execute

Bravo (Uniswap)
- Cancel 
- Queue 
- Execute


src/app/proposals/components/AgoraGovCancel.tsx
```diff
@@ -98,9 +87,8 @@ export const ProposalCancelButton = ({ proposal }: Props) => {
           )}
 
           <TooltipContent>
-            <div className="flex flex-col gap-1 p-2">
-              <div>Only the admin wallet can cancel proposals:</div>
-              <div className="font-semibold">{adminAddress}</div>
+            <div className="flex flex-col p-2">
+              <div>{"You don't have permission to cancel this proposal."}</div>
```
mcgingras, line 91: yesss love this pattern for getting around the unescaped entities

src/app/proposals/components/BravoGovQueue.tsx
```diff
@@ -0,0 +1,51 @@
...
+    if (isError) {
+      toast.error(`Error queuing proposal ${error?.message}`, {
+        duration: 5000,
+      });
+    }
+  }, [isSuccess, isError, error]);
+
+  return (
+    <>
+      {!isFetched && (
```
mcgingras, line 44: I think if we queue a tx, then refresh the page, we will lose the local state for data.hash, which means isFetched will be false again even though the tx was kicked off and possibly successful. The only way around this would be indexing the data (as far as I know). In the interest of time, I think it's okay, but maybe we flag it with a comment or something.

(I think this is an edge case but didn't test)

src/app/proposals/components/OZGovExecute.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const [canExecute, setCanExecute] = useState(false);
+  const [executeTime, setExecuteTime] = useState(new Date());
+
+  // Check whether user has the EXECUTOR_ROLE
+  const { data: hasExecuteRole } = useContractRead({
+    address: contracts.timelock!.address as `0x${string}`,
+    abi: contracts.timelock!.abi,
+    functionName: "hasRole",
+    args: [
+      "0xd8aa0f3194971a2a116679f7c2090f6939c8d4e01a2a8d7e41d55e5351469e63", // EXECUTOR_ROLE
```
mcgingras, line 38: Is this the same for all OZ gov contracts? Maybe we could move this into a const

src/lib/proposalUtils.ts
```diff
@@ -717,6 +717,20 @@ export async function getProposalStatus(
   return "QUEUED";
 }
 
+export const proposalToCallArgs = (proposal: Proposal) => {
+  const dynamicProposalType: keyof ParsedProposalData =
```
mcgingras, line 721: just a comment that I found our ParsedProposalData object hard to work with. What about you? Maybe we could refactor it at some point.

src/app/proposals/components/ProposalStateAdmin.tsx
```diff
@@ -2,20 +2,26 @@
```
mcgingras, line 2: really nice component -- I like this organization

mcgingras (approved):

src/app/proposals/components/OZGovExecute.tsx
andreitr: I am not sure about the other contracts. ENS modified their version a bit. Good call on making this a constant though

PR #554, opened 2024-10-11
Kent/agora 2553 spike an orchestrator testing pattern

jefag:


tests/orchestrator/setEnvVars.ts
```diff
@@ -0,0 +1,5 @@
+export function setEnv(envConfig: { [key: string]: string }) {
+  Object.keys(envConfig).forEach((key) => {
+    process.env[key] = envConfig[key];
+  });
+}
```
jefag, line 5: Is this being called somewhere? What is invoking this? I don't see it.

tests/orchestrator/orchestrator.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    return { tenant: tenant.name, status: "passed" };
+  } catch (error) {
+    return { tenant: tenant.name, status: "failed" };
+  }
+}
+
+// Main loop for test runner
+async function main() {
+  const config: Config = loadConfig() as Config;
```
jefag, line 65: I would have expected a `tenant-config.yaml` to load a single tenant's configuration.  Not a set of tenants.  

The loop should be over a set of tenant-config files, for instance all the files in a directory like this one: https://github.com/voteagora/tenants/tree/main/envs/prod

tests/orchestrator/setEnvVars.ts
jefag: FWIW, I'm not asking because I'm worried about about dead code.  I'm asking, because a key part of this is how we're managing the sets of environment variables in the orchestrator.

PR #559, opened 2024-10-12
Ensure isBadgeholder on ballot submission

mdehoog:
The ballet submission address is not validated to be a badgeholder, because the `isBadgeholder` variable isn't extracted from the `fetchBadgeholder`. This allows anyone with a valid signature to submit ballots.


stepandel (approved):
thank you!

PR #572, opened 2024-10-23
Hide Proposal Cancel for non admins

andreitr:
https://linear.app/agora-app/issue/AGORA-3106/proposal-cancel-admin-only

image


src/app/proposals/components/ProposalStateAdmin.tsx
```diff
@@ -45,9 +58,9 @@ export const ProposalStateAdmin = ({ proposal }: Props) => {
     switch (proposal.status) {
       case PROPOSAL_STATUS.ACTIVE:
       case PROPOSAL_STATUS.PENDING:
-        return "This proposal still can be cancelled";
+        return "This proposal still can be cancelled by the admin.";
```
mcgingras, line 61: "This proposal can still be cancelled by the admin" reads better to me, wdyt?

mcgingras (approved):

src/app/proposals/components/ProposalStateAdmin.tsx
andreitr: Reads better to me as well! I will update it.

PR #574, opened 2024-10-23
Tenant-level changes from the private branch 

andreitr:
- Typed toggle configs 
- Expanded endorsed config in the tenant


mcgingras (approved):
👍

PR #581, opened 2024-10-28
Fixing OP delegation

andreitr:
Delegation values were cached during component initialization. This PR ensures that the values are refreshed with every update.


mcgingras (approved):
ahh these query keys can be tricky! I did something similar on the claim flow

PR #590, opened 2024-11-04
Gasless voting

stepandel:


src/hooks/useSponsoredVoting.tsx
```diff
@@ -0,0 +1,125 @@
...
+  >(undefined);
+
+  const write = useCallback(() => {
+    const _sponsoredVote = async () => {
+      // Sign the vote
+      setWaitingForSignature(true);
+      try {
+        const signature = await signTypedDataAsync({
+          domain: {
+            name: "ENS Governor",
```
stepandel, line 44: note: this name should be dynamic -- this will only work for ENS

src/lib/tenant/configs/ui/uniswap.ts
```diff
@@ -218,6 +218,22 @@ export const uniswapTenantUIConfig = new TenantUI({
       name: "info/governance-charts",
       enabled: true,
     },
+    {
+      name: "sponsoredVote",
+      enabled: true,
+      config: {
+        domain: "Uniswap Governor Bravo",
```
andreitr, line 225: We will need to update this to handle different contract names between Sepolia and Mainnet.

src/lib/tenant/configs/ui/ens.ts
```diff
@@ -87,6 +87,22 @@ export const ensTenantUIConfig = new TenantUI({
...
+      config: {
+        domain: "ENS Governor",
+        version: "1",
+      },
+    },
+    {
+      name: "sponsoredDelegate",
+      enabled: true,
+      config: {
+        domain: "Ethereum Name Service",
```
andreitr, line 102: We will need to update this to handle different contract names between Sepolia and Mainnet.

src/lib/contracts/common/interfaces/IMembershipContract.ts
```diff
@@ -21,4 +21,6 @@ export interface IMembershipContract extends BaseContract {
     [bigint],
     "view"
   >;
+
+  nonces?: TypedContractMethod<[owner: AddressLike], [bigint], "view">;
```
andreitr, line 25: I added nonces as an optional method here just to prevent TypeScript from complaining. I have no idea if ERC1155 tokens implement EIP-2612. As of right now, this should never get called.

src/app/api/v1/relay/vote/route.ts
```diff
@@ -0,0 +1,35 @@
...
+export async function POST(request: NextRequest) {
+  const authResponse = await authenticateApiUser(request);
+
+  if (!authResponse.authenticated) {
+    return new Response(authResponse.failReason, { status: 401 });
+  }
+
+  try {
+    const body = await request.json();
+    console.log("body", body);
```
mcgingras, line 21: console.log here

src/hooks/useSponsoredDelegation.ts
```diff
@@ -0,0 +1,78 @@
...
+  const call = async () => {
+    setIsFetching(true);
+    setIsFetched(false);
+
+    const latestBlock = await contracts.token.provider.getBlock("latest");
+    const expiry = (latestBlock?.timestamp || 0) + 1000;
+
+    const signature = await signTypedDataAsync({
+      domain: {
+        name: gasRelayConfig.domain,
```
mcgingras, line 49: I feel like this shouldn't be part of the config -- it's always going to be the name of the token, right? So we can just read the `name` property off of the token contract, then use that? Addresses the comment you left in the config file itself

mcgingras (approved):

src/hooks/useSponsoredDelegation.ts
andreitr: brilliant. I wish I had thought about that 🤦

src/hooks/useGovernorName.ts
mcgingras, line 1: @andreitr this is an empty file right now?

src/components/Dialogs/UndelegateDialog/UndelegateDialog.tsx
```diff
@@ -46,25 +47,40 @@ export function UndelegateDialog({
...
   const { data: delegateeEnsName } = useEnsName({
     chainId: 1,
     address: delegatee?.delegatee as `0x${string}`,
   });
 
-  const { isError, writeContract: write, data } = useWriteContract();
+  const {
+    call,
+    isFetching: isProcessingSponsoredDelegation,
+    isFetched: didProcessSponsoredDelegation,
```
mcgingras, line 60: NIT and not blocking but might be nice to change these to `isProcessingSponsoredUnDelegation`

mcgingras (approved):

src/hooks/useGovernorName.ts
```diff
@@ -0,0 +1,22 @@
...
+  const { data, isFetching, isFetched } = useQuery({
+    queryKey: [GOVERNOR_NAME_QK, contracts.governor.address],
+    queryFn: async () => {
+      return await contracts.governor.contract.name();
+    },
+    enabled: enabled,
+  });
+
+  return { data, isFetching, isFetched };
+};
```
stepandel, line 22: I feel like this is a bit unnecessary. The contract name will never going to change even for upgradable contracts. It might be better to extend the contract config to include the contract name and pull it from there

andreitr: @stepandel In the case of Uniswap and ENS, the contract names differ between testnet and mainnet. This just makes it easier. Also, we don’t have to keep track of those names in the config. My vote is to keep this here.

src/lib/tenant/tenantUI.ts
```diff
@@ -16,10 +16,7 @@ export type UIEndorsedConfig = {
 };
 
 export type UIGasRelayConfig = {
-  domain: {
-    name: string;
-    version?: string;
-  };
+  version?: string;
```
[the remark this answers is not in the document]
andreitr: @mcgingras this has to do with how EIP-2612 is implemented. Uniswap Bravo uses a custom implementation that does not use version in the encoding, where ENS (OZ implementation) does.

PR #591, opened 2024-11-07
Andrei/agora 3162 scw implement delegation voting

andreitr:
image


src/app/api/common/quorum/getQuorum.ts
```diff
@@ -37,7 +37,10 @@ async function getQuorumForProposal(proposal: ProposalPayload) {
 
       // If no quorum is set, calculate it based on votable supply
       if (!quorum) {
-        votableSupply = await findVotableSupply({ namespace });
+        votableSupply = await findVotableSupply({
+          namespace,
+          address: contracts.token.address,
```
mcgingras, line 42: do these have to be lowercased? Or maybe we auto lowercase at the `findVotableSupply` level so its not necessary to pass in each time

src/app/api/paymaster/fetchPaymasterData.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  const response = await fetch(ui.smartAccountConfig!.paymasterUrl, {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({
+      secret: process.env.PAYMASTER_SECRET,
+      userOp: params,
```
mcgingras, line 19: not super big deal, but could be nice to use zod to validate the params we are passing in here

src/hooks/useSmartAccountDerive.ts
```diff
@@ -0,0 +1,165 @@
...
+});
+
+const lyraBundlerClient = createBundlerClient({
+  chain: contracts.token.chain,
+  transport: combinedTransport,
+  cacheTime: 1000,
+});
+
+const bundlerTransport = http(ui.smartAccountConfig?.bundlerUrl);
+const nodeTransport = http(
```
mcgingras, line 72: Is the conduit rpc just because derive is on a specific chain that no other providers support?

src/hooks/useSmartAccountDerive.ts
```diff
@@ -0,0 +1,165 @@
+import {
```
mcgingras, line 1: looks good, I think, most of this was over my head wrt the actual smart contract stuff. For the future but I wonder if we could pass in some params like which chainId it's for and make it compatible with any tenant rather than just derive.

mcgingras (approved):

src/app/api/common/delegates/getDelegateForSCW.ts
```diff
@@ -0,0 +1,18 @@
+import "server-only";
+
+import prisma from "@/app/lib/prisma";
+import { cache } from "react";
+import Tenant from "@/lib/tenant/tenant";
+
+// Returns an owner delegate for a given SCW address
+async function getDelegateForSCW(address: string) {
```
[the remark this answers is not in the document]
andreitr: So this is the reverse lookup of delegate by the scw wallet address. The scw addresses will not have delegate statements.

src/app/api/common/quorum/getQuorum.ts
andreitr: yes, the addresses are lowercase when they are returned from the TenantContract - the address is a getter function that transforms the value.

src/app/api/paymaster/fetchPaymasterData.ts
andreitr: good call. I need to start using zod more!

src/hooks/useSmartAccountDerive.ts
andreitr: correct

src/hooks/useSmartAccountDerive.ts
```diff
@@ -0,0 +1,165 @@
...
+} from "@alchemy/aa-core";
+
+import { createLightAccount, LightAccount } from "@alchemy/aa-accounts";
+import { useEffect, useState } from "react";
+import Tenant from "@/lib/tenant/tenant";
+import { fetchPaymasterData } from "@/app/api/paymaster/fetchPaymasterData";
+
+const ZERO_ADDRESS = "0x0000000000000000000000000000000000000000";
+const DUMB_SIGNATURE =
+  "0xfffffffffffffffffffffffffffffff0000000000000000000000000000000000aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaabb1c";
```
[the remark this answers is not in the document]
andreitr: The value is a placeholder but we have to have it when we are intiating a paymaster call. I

PR #594, opened 2024-11-07
Upgrade prisma from 4.16.2 to 5.22

jefag:

PR #598, opened 2024-11-09
Upgrade prisma from 4.16.2 to 5.22

ilamanov:
When I deploy on my own Vercel instance, everything works now!

A few side notes
• For my own deployment, I had to add `yarn run generate-typechain & npx prisma generate` to the `build` step in package.json. Since you guys don't have it, I assumed you use a different mechanism to generate these things. Or they are cached (in which case it's still a good idea to regenerate them on every build - I can add these scripts back if you want)
• In https://github.com/voteagora/agora-next/pull/594, I saw that an external `decimal.js` library was used for `Decimal`. Prisma v5 still has `Decimal`. It just moved to a different import - https://www.prisma.io/docs/orm/more/upgrade-guides/upgrading-versions/upgrading-to-prisma-5#using-public-apis-from-prismaclientruntime. So I used this one instead of an external lib


jefag:
Great stuff @ilamanov !  And amazing turn-around time!

jefag:
# RE: `Decimal`

tl;dr ✅ LGTM

`Decimal` this is great! I had seen those docs, but on my local that wasn't working for some reason.  I couldn't get both the `Decimal` Type and Constructor to work from prisma 5, until I swapped out the library for the independent one.  🤔  But, really happy to avoid the extra lib, I'm sure I can sort out my local.

jefag:
# RE: Build scripts

tl;dr ✅ 

We have `npx prisma generate && yarn generate-typechain && next build` written in vercel build command.  Happy to re-evaluate how we do this with the team with your input.  I didn't have to do anything in vercel (except approve) in order to get those preview links above.

jefag:
# RE: prismaUtils.ts

This does solving the type-union issue.  I'm not crazy about this approach here, because it means one more thing to go manually touch in the code-base every time we add a new Tenant.  But if it gets us onto prisma 5 by Monday AM, I ok with it.  There is a silver-lining here, as it ring-fences the growth of this pattern, and gives us an organized way to deprecate that approach, in favor of raw queries, which is on our road-map anyway.

jefag:
@ilamanov just need to run prettier, then I'd like to merge.

This step should pass:

image


...the rest can fail.

ilamanov:
> @ilamanov just need to run prettier, then I'd like to merge.
> 
> This step should pass:
> 
> image
> ...the rest can fail.

Great! Fixed this

PR #600, opened 2024-11-11
Always show 0% participation

ilamanov:
Bug - Delegates who have NEVER voted, have null participation, rather than 0% participation.

Fixes: https://linear.app/agora-app/issue/AGORA-3235/bug-delegates-who-have-never-voted-have-null-participation-rather-than

### DIFF
--- a/src/components/Delegates/DelegateCardList/DelegateCard.tsx
+++ b/src/components/Delegates/DelegateCardList/DelegateCard.tsx
@@ -54,11 +54,9 @@ const DelegateCard = ({
               <span className="text-primary font-bold">
                 {formatNumber(delegate.votingPower.total)} {token.symbol}
               </span>
-              {votingStats && (
-                <span className="text-primary font-bold">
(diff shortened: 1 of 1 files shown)


src/components/Delegates/DelegateCardList/DelegateCard.tsx
```diff
@@ -54,11 +54,9 @@ const DelegateCard = ({
...
                 {formatNumber(delegate.votingPower.total)} {token.symbol}
               </span>
-              {votingStats && (
-                <span className="text-primary font-bold">
-                  {(votingStats?.last_10_props || 0) * 10}% Participation
-                </span>
-              )}
+              <span className="text-primary font-bold">
+                {(votingStats?.last_10_props || 0) * 10}% Participation
+              </span>
```
mcgingras, line 59: one problem is that while the data is loading it shows "0% participation" -- it doesn't look that bad, I pulled it down and ran it. Might be confusing though

ilamanov:
Superseded by https://github.com/voteagora/agora-next/pull/607

PR #607, opened 2024-11-14
Always show 0% participation

ilamanov:
Bug - Delegates who have NEVER voted, have null participation, rather than 0% participation.

Fixes: https://linear.app/agora-app/issue/AGORA-3235/bug-delegates-who-have-never-voted-have-null-participation-rather-than

Supersedes https://github.com/voteagora/agora-next/pull/600
Two improvements on top of the old PR:
• Pushed from voteagor repo instead of my own fork
• FIxes "one problem is that while the data is loading it shows "0% participation" -- it doesn't look that bad, I pulled it down and ran it. Might be confusing though" (https://github.com/voteagora/agora-next/pull/600#discussion_r1840895551)

### DIFF
--- a/src/components/Delegates/DelegateCardList/DelegateCard.tsx
+++ b/src/components/Delegates/DelegateCardList/DelegateCard.tsx
@@ -23,6 +23,7 @@ const DelegateCard = ({
 }) => {
   const { token } = Tenant.current();
   const { advancedDelegators } = useConnectedDelegate();
+
   const { data: votingStats, isPending: isVotingStatsPending } = useVotingStats(
     {
       address: delegate.address as `0x${string}`,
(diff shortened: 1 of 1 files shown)


ilamanov:
The bug fix related to not showing anything when results are loading has not been verified yet because I don't have uniswap data locally. Will test once I have it or Vercel previews work

ilamanov:
> The bug fix related to not showing anything when results are loading has not been verified yet because I don't have uniswap data locally. Will test once I have it or Vercel previews work

Just verified that it works

https://github.com/user-attachments/assets/34cefcff-28f3-43b7-b8b4-180ac73647d1



https://github.com/user-attachments/assets/34cefcff-28f3-43b7-b8b4-180ac73647d1

jefag:
@mcgingras wdyt?

ilamanov:
Bump on this @jefag @mdehoog

PR #610, opened 2024-11-18
Fix voter card loading

ilamanov:
Fixes https://linear.app/agora-app/issue/AGORA-3240/troubleshoot-fix-hover-card-that-never-loads

For the default message (when the statement is `null`), I just show `No statement` as in the screen recording below. But I'm not sure if there is a better message we can display here. All the content of a valid hover card seem to be dependent on `statement` being non-null so there is really nothing we can show as a fallback.

### Demo
https://github.com/user-attachments/assets/724d1f22-6843-4b49-9cf4-38a6cf614912

### DIFF
--- a/src/components/Votes/VoterHoverCard.tsx
+++ b/src/components/Votes/VoterHoverCard.tsx
@@ -50,12 +50,10 @@ export default function VoterHoverCard({
     );
   }
 
-  if (delegate.statement === null) {
-    return <p className="text-center">No statement</p>;
-  }
-
   const truncatedStatement =
-    delegate.statement.payload?.delegateStatement?.slice(0, 120) || "";
+    delegate.statement === null
+      ? null
+      : delegate.statement.payload?.delegateStatement?.slice(0, 120) || "";
 
   return (
     <VStack gap={4} className="h-full w-[300px]">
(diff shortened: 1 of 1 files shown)


jefag:
Glad it's not actually a real bug.  Just unfinished polish.

So there really is two cases.

# ENS Name exists, with / without a profile pic from ENS:

image

# No ENS Name exists (and therefore there would be no profile pic)

image

I think we still want the "delegate" button, because that's still technically possible. You might want to delegate to an anon, it's wierd, but its valid.

ilamanov:
> Glad it's not actually a real bug. Just unfinished polish.
> 
> So there really is two cases.
> 
> # ENS Name exists, with / without a profile pic from ENS:
> image
> # No ENS Name exists (and therefore there would be no profile pic)
> image
> I think we still want the "delegate" button, because that's still technically possible. You might want to delegate to an anon, it's wierd, but its valid.

Thanks, fixed now!

https://github.com/user-attachments/assets/d9bb7fb6-7e85-4d58-8a22-bc4dd41f3203

I essentially kept the same rendering of the card as in the normal case (when the statement is available). Just changed the statement to "This delegate has not yet created a delegate statement."

mcgingras (approved):
This is great, I agree with the back and forth between you and Jeff. 
Thanks!

PR #612, opened 2024-11-18
Adds CSS route to get tenant CSS

mcgingras:


jefag (approved):
Sweeeeeeeet!

But, since we're going to glue this to other stuff, and and it's key to our open-source strategy...can we make the endpoint something without the word `tenant`? 

Eg. `style.css`, `branding.css`, `custom.css` or something else?  

From the perspective of somebody deploying this themselves, they wouldn't think of themselves as a tenant.  That's all.

PR #616, opened 2024-11-21
Delegate table

ilamanov:
### Michael's latest note:
See my comment trail at the bottom of this commit log, starting [here](https://github.com/voteagora/agora-next/pull/616#issuecomment-2603166316).

The only remaining issue I can find is the "delegates => citizens" issue. I'm not sure how big of an issue this is, since it's a pretty edge-y edge case. I spent about 15 minutes trying to debug and didn't get anywhere. I'd estimate a 1/2 day to a day worth of time to actually re-architect the code such that this bug is fixed. I think it's probably a better decision for product to evaluate if the tradeoff is worth the time, since it's such an edge case.

---

Adds a table view for delegates (https://linear.app/agora-app/issue/AGORA-3164/as-a-user-i-can-select-between-delegate-card-views)

### Demo

https://github.com/user-attachments/assets/c92a1e6e-b1a0-420a-89e1-6de05e51a04c

### TODO
One thing that has not yet been implemented is the `For/Against/Abstain` column. I didn't implement it because this data doesn't exist in the current request. See my inline comment on `getDelegates.ts`


src/app/api/common/delegates/getDelegates.ts
```diff
@@ -304,6 +304,7 @@ async function getDelegates({
       },
       citizen: delegate.citizen,
       statement: delegate.statement,
+      numOfDelegators: BigInt(delegate.num_of_delegators),
```
ilamanov, line 307: So, the data for `Delegated from` column exists here. We just need to propagate it.

But data for the `For/Against/Abstain` column does not exist. It only exists as part of the `getDelegate` request function in the same file. But for this function (`getDelegates`) it doesn't exist. Would someone be able to add `votedFor`, `votedAgainst`, etc as in the `getDelegate` function below? (And then I can sync to main and build my PR on top of that). Happy to do it too if I can get instructions. I tried to understand the code myself but the SQL here is way too complex and I'm afraid to break something

src/app/delegates/page.jsx
```diff
@@ -39,7 +40,9 @@ export default async function Page({ searchParams }) {
     <section>
       <Hero />
       <Suspense fallback={<DelegateCardLoadingState />}>
-        <DelegateCardWrapper searchParams={searchParams} />
+        <LayoutProvider>
```
ilamanov, line 43: We need this LayoutContext to share the state for the picked layout between two sibling components. If we were to do it in a more straightforward way where we just store this state in the parent and pass to its children, we would need to refactor the parent (`DelegateCardWrapper.tsx`) to make it a client component which means moving all async functions one level up. This involves a lot of movement of async functions around. 

I thought this React context approach is simpler in this case. It also benefits us because this layout preference (grid vs table) may be used in other components in the future as well as I think this won't be limited to the Delegates view only. In that case we can just reuse this context without trying to pass the state from the parent.

If there is a strong preference to avoid this React context, I'm happy to turn DelegateCardWrapper into a client component. But like I mentioned, there will be a considerable refactor

src/components/Delegates/DelegateCardList/DelegateContent.tsx
ilamanov, line 1: Component that is responsible for rendering delegates (whether grid or table view)

src/components/Delegates/DelegatesTabs/DelegatesTabs.tsx
```diff
@@ -50,7 +58,7 @@ export default function DelegateTabs({ children }: { children: ReactNode }) {
   return (
     <Tabs
       className="max-w-full"
-      defaultValue={tabParam || "delegates"}
```
ilamanov, line 53: I extracted the internal state of `Tabs` into a managed state in this (`DelegateTabs`) component because we need access to that state in this component. We render the layout picker based on its state (only when value = `delegates`)

src/components/Delegates/DelegatesTabs/DelegatesTabs.tsx
```diff
@@ -5,15 +5,24 @@ import { useRouter, useSearchParams } from "next/navigation";
...
 import { useAddSearchParam, useDeleteSearchParam } from "@/hooks";
 import StakeholdersFilter from "@/app/delegates/components/StakeholdersFilter";
 import IssuesFilter from "@/app/delegates/components/IssuesFilter";
 import EndorsedFilter from "@/app/delegates/components/EndorsedFilter";
 import DelegateeFilter from "@/app/delegates/components/DelegatorFilter";
+import { LayoutGrid, AlignJustify } from "lucide-react";
+import { useLayout } from "@/contexts/LayoutContext";
 
 export default function DelegateTabs({ children }: { children: ReactNode }) {
+  const searchParams = useSearchParams();
```
ilamanov, line 19: Just moving these 2 lines up because they will be used in `useState`

src/contexts/LayoutContext.tsx
ilamanov, line 1: Context for managing the picked layout

ilamanov:
Commit above^: removed the type temporarily because it was giving preview build errors. Once we are settled on how to augment the data for `getDelegates` will properly add the type back

src/app/delegates/page.jsx
mcgingras: I don't mind the use of context here -- I agree it would be cumbersome to have to move all of the async calls out.

jefag:
I couldn't get uniswap to build.

https://vercel.com/voteagora/agora-next-uniswap/HeBEDhoYoKbeDeLtLZziUYqG7GY4?filter=errors

image

ilamanov:
> I couldn't get uniswap to build.
> 
> https://vercel.com/voteagora/agora-next-uniswap/HeBEDhoYoKbeDeLtLZziUYqG7GY4?filter=errors
> 
> image

I think we caught a weird intermediate state where my last commit wasn't propagated to Uniswap specifically. This is exactly the error I tried to avoid with my (now) second-to-last commit. I just pushed some trivial change again to re-trigger the preview build and now Uniswap builds successfully

Although `CI / Swagger Editor Validator Remote` fails. Seems to be unrelated to this PR. Does anyone know the cause of it?
CleanShot 2024-11-21 at 19 11 17

ilamanov:
Also, please let me know whether you want me to add the data for `For/Against/Abstain` (this is a slower option) or someone wants to do it in a separate PR and then I build on top of it (faster option)

https://github.com/voteagora/agora-next/pull/616#discussion_r1851448075

ilamanov:
> > I couldn't get uniswap to build.
> > https://vercel.com/voteagora/agora-next-uniswap/HeBEDhoYoKbeDeLtLZziUYqG7GY4?filter=errors
> > image
> 
> I think we caught a weird intermediate state where my last commit wasn't propagated to Uniswap specifically. This is exactly the error I tried to avoid with my (now) second-to-last commit. I just pushed some trivial change again to re-trigger the preview build and now Uniswap builds successfully
> 
> Although `CI / Swagger Editor Validator Remote` fails. Seems to be unrelated to this PR. Does anyone know the cause of it? CleanShot 2024-11-21 at 19 11 17

It's failing for another of my PRs: https://github.com/voteagora/agora-next/pull/610
So I am assuming it's a more general issue at the moment, not related to this PR specifically

jefag:
I'm looking into the busted build over on #618

mcgingras:
Screenshot 2024-11-26 at 10 30 25 AM

Nit, but can we remove the inner shadow on the de-emphesized state of the grid icon?

src/components/Delegates/DelegateCardList/DelegateContent.tsx
```diff
@@ -0,0 +1,42 @@
...
+export default function DelegateContent({
+  initialDelegates,
+  fetchDelegates,
+  isDelegatesCitizensFetching,
+  fetchDelegators,
+}: Props) {
+  const { layout } = useLayout();
+
+  return layout === "grid" ? (
+    <DelegateCardList
```
mcgingras, line 29: https://github.com/user-attachments/assets/239354db-c183-42a5-83e2-f4100d8dab04

seeing an issue -- maybe in this component? 
I get the app to crash if I start on list view -> go to citizens tab -> back to list view

src/components/Delegates/DelegateCardList/DelegateTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+      setDelegates(delegates.concat(data.data));
+      setMeta(data.meta);
+      fetching.current = false;
+    }
+  };
+
+  return (
+    <DialogProvider>
+      <Table className="min-w-full">
+        <TableHeader className="text-xs text-secondary sticky top-0 bg-white z-10">
```
mcgingras, line 59: some design tweaks
Screenshot 2024-11-26 at 10 41 13 AM

src/components/Delegates/DelegateCardList/DelegateTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+          <TableRow>
```
mcgingras, line 60: it'd be nice if clicking the row took you to the delegate profile, the same way clicking the card takes you to delegate profile in grid view

src/components/Delegates/DelegatesTabs/DelegatesTabs.tsx
```diff
@@ -73,6 +81,20 @@ export default function DelegateTabs({ children }: { children: ReactNode }) {
...
+              <button onClick={() => setLayout("grid")}>
+                <LayoutGrid
+                  className={`h-6 w-6 ${layout === "grid" ? "text-secondary fill-secondary" : "text-secondary/30 fill-secondary/30"}`}
+                />
+              </button>
+              <button onClick={() => setLayout("list")}>
+                <AlignJustify
+                  className={`h-6 w-6 ${layout === "list" ? "text-primary" : "text-secondary/30"}`}
+                />
+              </button>
```
mcgingras, line 95: I'm not familiar with these components, I'd prefer if we could write regular html elements and style them with tailwind. I don't really like utility components like <AlignJustify /> which I'm guessing just positions content a certain way. Something we could do with tailwind.

We had something similar `<HStack />` and `<VStack/>` which only rendered a div with flex and flex-col or flex-row and I think it ends up being more challenging when you have another layer of abstraction via a component that is only adding a few classes.

Sorry to be a pest on this!

mcgingras (changes requested):
left some comments!

ilamanov:
> Screenshot 2024-11-26 at 10 30 25 AM
> Nit, but can we remove the inner shadow on the de-emphesized state of the grid icon?

Done

src/components/Delegates/DelegateCardList/DelegateContent.tsx
ilamanov: For some reason the video is no longer loading. Tried to repro this but I can't find the "Citizens" tab on ENS. Which chain is it available on?

mcgingras:
As noted by Nazar, the data returned does not include for, against, abstain. We don't include this on the cards that we toggle between for list <> card, so I feel like it's okay to take this out completely. It's better than showing false data (all 0's) for every user. Now, the data shown on the cards is consistent with the data shown in the table.

mcgingras (approved):

mcgingras:
### Steps to test

- Go to any deployment
- Go to voters
- Click the "table" icon on the right side of the page by the filters

Expectation: the table renders correctly, the table renders correct data, the UI is consistent with other colors provided by the tenant config.

- Click the "card" icon to go back to card view

Expectation: the card view should render again, the card view should render correct data.

andreitr (approved):
LGTM

bryan-agora:
Testing Results:
Quick note across all tenants - the initial page load of the Voters tab was painfully slow. I didn't capture this on Loom, but it was so slow that it felt broken.

Cyber - Good
Boost - Good
Derive:
- Styling for the Delegate List off
- It was also hopping back and forth between List and Card view without any prompting
- Loom: https://www.loom.com/share/474e08febaa54fe4a277200a92b6c360 (sorry my mic was bad, might need to mute but you can see what I'm talking about)
Etherfi - Also jumping between List and Card view
OP - Also jumping between List and Card view
PG - Good
Scroll - Also jumping between List and Card view
Uniswap - Also jumping between List and Card view

bryan-agora:
Updated testing looks good on my end! QA Approved!

PR #617, opened 2024-11-21
OP executable proposals: required app upgrades

mcgingras:
## Prepares app for OP executable proposals test
1. replaces PTC testnet address
2. pays down tech debt to allow tenant config to define which voting modules it supports, and the addresses of each of those modules. Previously, the addresses of the modules were hardcoded, and we were supporting only OP mainnet and cyber testnet (???). Now these are configurable at the tenant level, and will throw errors if not configured.

Since 2) changes PLM code, it would be great if we could get a few QA passes done. All we really need to test is that we can create proposals for each supposed voting module on each tenant. After the proposal is made, it's a different code path that these changes don't touch.


src/app/proposals/draft/[id]/page.tsx
```diff
@@ -74,14 +74,14 @@ export default async function DraftProposalPage({
           </section>
           <section className="col-span-1">
             <div className="bg-wash border border-line rounded-2xl p-4">
-              <p className="mt-2">
+              <div className="mt-2">
```
Comment (mcgingras) [inline src/app/proposals/draft/[id]/page.tsx:77]: React was complaining about nesting a dev in a p tag... simple to change, doesn't matter much

src/lib/tenant/configs/ui/ens.ts
```diff
@@ -132,7 +132,18 @@ export const ensTenantUIConfig = new TenantUI({
...
+            type: ProposalType.BASIC,
+            prodAddress: null,
+            testnetAddress: null,
+          },
+          {
+            type: ProposalType.SOCIAL,
+            prodAddress: null,
+            testnetAddress: null,
+          },
+        ],
```
mcgingras, line 146: basic and social type do not need to define any address -- there is no module to support for basic or social

src/components/Proposals/ProposalPage/OPProposalPage/OPProposalOptimisticPage.jsx
```diff
@@ -78,16 +77,13 @@ export default async function OPProposalPage({ proposal }) {
...
-    // 2 Colum Layout: Description on left w/ transactions and Votes / voting on the right
-    <HStack
-      gap={16}
-      justifyContent="justify-between"
-      alignItems="items-start"
-      className="max-w-[76rem] flex-col sm:flex-row items-stretch sm:items-start justify-end sm:justify-between"
-    >
-      <ProposalDescription proposalVotes={proposalVotes} proposal={proposal} />
-      <div>
-        <StandardProposalDelete proposal={proposal} />
```
andreitr, line 90: Removed StandardProposalDelete

andreitr (approved):
Looks LGTM

PR #615, opened 2024-11-20
Mobile Footer Tweaks

kent:

PR #625, opened 2024-11-26
Improved footer

ilamanov:
Implements https://linear.app/agora-app/issue/AGORA-3201/fix-floating-footer#comment-2b29072d (and also fixes the underlying floating footer issue which was built on top of https://github.com/voteagora/agora-next/pull/615/files)

### Demo 
The UI of the drawer is a bit different from the screenshot in Yitong's comment above but I feel like the right component to use here is a Drawer component. And I just used the one from shadcn. Let me know if there is a strong preference to use the UI from the screenshot instead.

https://github.com/user-attachments/assets/9bfbe21f-cdc9-4481-aa8d-ff489522fd6d

### TODO
I'm also assuming that other links now need to be activated? It's weird to only show the changelog link. I see that you links are provided in the TenantUI constractor. Which chains/protocols should I add additional links to? All of them?

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
     "tailwindcss-animate": "^1.0.6",
     "utf-8-validate": "^6.0.3",
     "uuid": "^9.0.1",
+    "vaul": "^1.1.1",
     "viem": "2.21.16",
     "wagmi": "^2.12.17",
     "zod": "^3.22.4"
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -2,7 +2,7 @@ import React from "react";
 
 const Footer: React.FC = () => {
   return (
-    <footer className="text-secondary pt-4 pb-8 gap-1 flex w-full justify-center text-sm px-4">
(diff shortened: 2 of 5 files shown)


mcgingras:
Hi Nazar. Sorry that there was basically no design input on this ticket. I don't think the drawer works very well here. Taking a look at your demo it feels jarring for the footer to transition into a drawer, and is especially odd when were going through all of that UI work just to show a link. 

I think what would make this better is if instead of changing to a drawer, we hid bottom row by default, but then animated it into view. The vote tab on our proposals page is a good example of this:

https://github.com/user-attachments/assets/ffa8a972-f61f-4085-bc7b-54f8a65b63c7

What I would expect here is that instead of opening a drawer, we just animate the footer element so the second row pops into view. This way the element feels fluid -- it's just growing into space, and there isn't as much of a jarring transition into a totally new drawer element. Let me know if this makes sense, or if you have questions. Sorry there wasn't any design guidance here!

mcgingras (changes requested):
see comment!

PR #629, opened 2024-11-27
Fix OP admin voting settings issues

ilamanov:
Fixes both bugs reported in https://linear.app/agora-app/issue/AGORA-3309/bug-in-time-to-block-conversion-and-input-field-in-op-admin-voting :

If the user puts in 1 "hours", for "Voting Period", it broadcasts a message equal to 900, which is 30 minutes.  If the user puts in 1, it should broadcast 1800.

Additionally, in either field (Voting period or Voting delay).  The text box is awkward if you press backspace.  It won't let you clear the field.  The behaviour should be to allow the user to clear the field, in order to type a new number.  Perhaps gray out the button, if the field is empty.

### Demo

https://github.com/user-attachments/assets/c5335c6e-43e2-48a8-b000-3e5ca055fa14

### DIFF
--- a/src/components/Admin/GovernorSettings.tsx
+++ b/src/components/Admin/GovernorSettings.tsx
@@ -49,29 +49,37 @@ export default function GovernorSettings() {
   useEffect(() => {
     if (data) {
       setVotingPeriod(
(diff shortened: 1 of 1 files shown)


jefag:
BTW, great stuff, thank you @ilamanov !

PR #630, opened 2024-11-27
[Improvement] Hide Delegate Card Activity if Absolute # of proposals < 3

kent:
Was playing around with Boost and noticed that I was an inactive delegate even though there were no proposals.
Added a quick helper to fetch totalPrposals efficiently, then toggle based on that. 

Boost, no inactive delegare card, because less than 3 proposals.

CleanShot 2024-11-27 at 18 57 29@2x

ENS, Inactive card is there, because more than 3 proposals
CleanShot 2024-11-27 at 18 58 09@2x

OP, Inactive card is there, because more than 3 proposals
CleanShot 2024-11-27 at 18 59 53@2x

### DIFF
--- a/src/lib/prismaUtils.ts
+++ b/src/lib/prismaUtils.ts
@@ -460,3 +460,40 @@ export function findStakedDeposits({
       throw new Error(`Unknown namespace: ${namespace}`);
   }
 }
+
+export function getProposalsCount({
+  namespace,
+  contract,
+}: {
+  namespace: TenantNamespace;
+  contract: string;
+}) {
+  const condition = {
+    where: {
+      contract,
+    },
+  };
+
+  switch (namespace) {
+    case TENANT_NAMESPACES.OPTIMISM:
+      return prisma.optimismProposals.count(condition);
+    case TENANT_NAMESPACES.ENS:
+      return prisma.ensProposals.count(condition);
+    case TENANT_NAMESPACES.ETHERFI:
+      return prisma.etherfiProposals.count(condition);
+    case TENANT_NAMESPACES.UNISWAP:
+      return prisma.uniswapProposals.count(condition);
+    case TENANT_NAMESPACES.CYBER:
+      return prisma.cyberProposals.count(condition);
+    case TENANT_NAMESPACES.SCROLL:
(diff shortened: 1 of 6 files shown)


src/lib/prismaUtils.ts
```diff
@@ -460,3 +460,40 @@ export function findStakedDeposits({
...
+      return prisma.scrollProposals.count(condition);
+    case TENANT_NAMESPACES.DERIVE:
+      return prisma.deriveProposals.count(condition);
+    case TENANT_NAMESPACES.PGUILD:
+      return prisma.pguildProposals.count(condition);
+    case TENANT_NAMESPACES.BOOST:
+      return prisma.boostProposals.count(condition);
+    default:
+      throw new Error(`Unknown namespace: ${namespace}`);
+  }
```
jefag, line 498: We don't really want to be extending this prisma pattern.  It's not adding enough value to keep using prisma like this.  It's better just to write the raw SQL, so it works for all deployments, rather than needing to copy lines for each new deployment.

src/lib/prismaUtils.ts
```diff
@@ -460,3 +460,40 @@ export function findStakedDeposits({
...
+}) {
+  const condition = {
+    where: {
+      contract,
+    },
+  };
+
+  switch (namespace) {
+    case TENANT_NAMESPACES.OPTIMISM:
+      return prisma.optimismProposals.count(condition);
```
jefag, line 479: And, regardless of if you keep the prisma based approach or switch to SQL, we need to only count proposals that are not cancelled and not snapshots, if we want to generalize this reliably.  Not mandatory here, but in the case of a new tenant, they could easily make 3 or 4, and then cancel all of them.

jefag (approved):
We could merge, but there are things that could be improved too.  Up to you.

src/lib/prismaUtils.ts
kent: This is great feedback @jefag, I am going to merge but fix in another PR. Ty

src/lib/prismaUtils.ts
kent: This is great feedback @jefag, I am going to merge but fix in another PR. Ty

PR #640, opened 2024-12-02
ProposalTypes fix (maybe)

andreitr:


src/lib/tenant/configs/ui/optimism.ts
```diff
@@ -250,17 +250,17 @@ export const optimismTenantUIConfig = new TenantUI({
         ],
         proposalTypes: [
           {
-            type: ProposalType?.BASIC,
+            type: ProposalType.BASIC,
```
mcgingras, line 253: If this fixes the issue, then no optional chaining is fine. But if we don't have the optional chaining then this blows up if we hit the undefined issue. So, rather than seeing empty voting module types in PLM for example, it would just crash the page

So... if this fixes works then I think we are okay! Otherwise, I'd keep these optional chaining `?` in

PR #650, opened 2024-12-05
Eliminate partial flag

jefag:
This PR fixes the delegation button for DAOs using partial delegation.

### DIFF
--- a/src/lib/prismaUtils.ts
+++ b/src/lib/prismaUtils.ts
@@ -45,13 +45,11 @@ export function findDelagatee({
 export function findAdvancedDelegatee({
   namespace,
   address,
-  contract,
-  partial,
+  contract
 }: {
   namespace: TenantNamespace;
   address: string;
   contract?: string;
-  partial?: boolean;
 }) {
   const condition = {
     where: {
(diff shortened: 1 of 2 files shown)


src/lib/tenant/configs/contracts/scroll.ts
```diff
@@ -75,6 +76,14 @@ export const scrollTenantContractConfig = ({
       provider,
     }),
 
+    proposalTypesConfigurator: new TenantContract<BaseContract>({
```
mcgingras, line 79: Nice ya, did the same in my branch here
https://github.com/voteagora/agora-next/pull/643

We will need a timelock too

src/lib/prismaUtils.ts
```diff
@@ -63,41 +61,23 @@ export function findAdvancedDelegatee({
...
-      return prisma.uniswapAdvancedDelegatees.findMany(
-        partial ? undefined : condition
-      );
+      return prisma.uniswapAdvancedDelegatees.findMany(condition);
     case TENANT_NAMESPACES.CYBER:
-      return prisma.cyberAdvancedDelegatees.findMany(
-        partial ? undefined : condition
-      );
+      return prisma.cyberAdvancedDelegatees.findMany(condition);
     case TENANT_NAMESPACES.SCROLL:
```
andreitr, line 73: I must have missed the time when we introduced this pattern. Why are we doing this?

jefag: It's because the Prisma upgrade, Prisma got more strict with their API.  No choice.  Plan is to kill off, using RAW SQL and/or DAONode.

PR #659, opened 2024-12-08
Gasless delegation

ilamanov:
This PR only adds the functionality to delegate by signature. The UI is not updated according to the designs yet. We check the ui toggle: if gasless delegation is enabled, we use `delegateBySig`, otherwise we use the regular `delegate` function

### DIFF
--- a/src/app/api/v1/relay/getRelayStatus.ts
+++ b/src/app/api/v1/relay/getRelayStatus.ts
@@ -3,7 +3,7 @@ import { cache } from "react";
 import { formatEther } from "viem";
 import { privateKeyToAccount } from "viem/accounts";
 
-const SPONSOR_PRIVATE_KEY = process.env.SPONSOR_PRIVATE_KEY;
+const SPONSOR_PRIVATE_KEY = process.env.NEXT_PUBLIC_GAS_SPONSOR_PK;
 const GAS_COST = 0.001108297;
 
 async function getRelayStatus() {
--- a/src/components/Dialogs/DelegateDialog/DelegateDialog.tsx
+++ b/src/components/Dialogs/DelegateDialog/DelegateDialog.tsx
@@ -3,6 +3,8 @@ import {
   useWriteContract,
   useEnsName,
   useWaitForTransactionReceipt,
+  useSignTypedData,
+  useReadContract,
 } from "wagmi";
 import { ArrowDownIcon } from "@heroicons/react/20/solid";
 import { Button } from "@/components/Button";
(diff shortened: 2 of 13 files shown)


src/app/api/v1/relay/getRelayStatus.ts
```diff
@@ -0,0 +1,26 @@
+import { getPublicClient } from "@/lib/viem";
+import { cache } from "react";
+import { formatEther } from "viem";
+import { privateKeyToAccount } from "viem/accounts";
+
+const SPONSOR_PRIVATE_KEY = process.env.SPONSOR_PRIVATE_KEY;
+const GAS_COST = 0.001108297;
```
mcgingras, line 7: is this for voting? maybe we can have a mapping of gas costs for different actions, since we are going to support a few (vote, delegate, maybe more in the future)

src/app/api/v1/relay/getRelayStatus.ts
```diff
@@ -0,0 +1,26 @@
...
+  const publicClient = getPublicClient();
+  const account = privateKeyToAccount(SPONSOR_PRIVATE_KEY as `0x${string}`);
+
+  const balance = await publicClient.getBalance({
+    address: account.address,
+  });
+
+  return {
+    balance: Number(formatEther(balance)),
+    remaining_votes: Math.floor(Number(formatEther(balance)) / GAS_COST),
```
mcgingras, line 22: same with here -- remaining_votes makes sense in the context of voting but how about delegations?

src/icons/infoRed.svg
mcgingras, line 1: is it possible for us to update icon component such that we can pass classnames down and have a single `info` icon that can be configured to any color?

src/lib/tenant/configs/ui/uniswap.ts
```diff
@@ -216,6 +216,10 @@ export const uniswapTenantUIConfig = new TenantUI({
       name: "info/governance-charts",
       enabled: true,
     },
+    {
```
mcgingras, line 219: this PR is for delegation though, right? Should we disable `sponsoredVote`?

src/components/Dialogs/DelegateDialog/DelegateDialog.tsx
```diff
@@ -57,16 +75,30 @@ export function DelegateDialog({
...
+    data: delegateTxHash,
+  } = useWriteContract();
 
   const {
     isLoading: isProcessingDelegation,
     isSuccess: didProcessDelegation,
     isError: didFailDelegation,
   } = useWaitForTransactionReceipt({
-    hash: data,
+    hash: ui.toggle("sponsoredDelegate") ? sponsoredVoteTxHash : delegateTxHash,
```
andreitr, line 89: Please create a separate dialog for gasless voting, perhaps named DelegateDialogGasless.tsx. This will make it easier for us to enable this component at the top level for specific tenants.

andreitr (changes requested):
Solid work, @ilamanov. I really love how clear your code is. A few minor suggestions + change request.

PR #671, opened 2024-12-13
small UI tweak for delete draft modal

yitongzhang:
**New:**
CleanShot 2024-12-12 at 16 05 40@2x

**Old**
CleanShot 2024-12-12 at 16 07 08@2x


jefag:
@yitongzhang did you test this on mobile?

PR #679, opened 2024-12-18
ENG-874 -  Ignore active proposals in % Participation

mcgingras:
Resolves issue where 10 concurrent proposals wipes out voter participation, since we are using past 10 proposals to calculate participation percentage. The fix here introduces a change to the way we calculate participation. We now only count your participation over the last ten proposals that have ended -- meaning the endBlock is in the past. 

In this PR we start migrating away from the voter_stats view, all participation based calculations now rely on a new query from agora-next - NOT a materialized view like voter_stats. We still need voter_stats for things like a count of for, against, abstain, but we may phase this view out over time since it is extremely slow.


andreitr (approved):
LGTM

PR #693, opened 2025-01-03
Uniswap quorum display fix

andreitr:
Only counting `For` votes for Uniswap quorum. 

Screenshot 2025-01-03 at 12 27 18 PM


jefag (approved):
Is this the only place where we're tripped up by this convention difference for Uniswap?  

Fine to merge, but would like to understand how this should be set in a governor/config based way, rather than a namespace toggle.

PR #696, opened 2025-01-08
Upgrade votable supply to be token-contract aware - derive only

jefag:
This PR upgrades `deriveVotableSupply` to be contract aware.

We need to do this for all other tenants still, but that can be safely handled in a future PR.


andreitr (approved):
LGTM

PR #701, opened 2025-01-09
Starting partial delegation feature branch for derive

jefag:


jefag:
Probably easier to just use the other branch as a feature branch.

https://github.com/voteagora/agora-next/pull/709

PR #705, opened 2025-01-09
Add Scroll Case to Pre-Snapshot Quorum

jefag:


jefag:
Discussed w/ @mcgingras before.  Confirmed on local.

PR #706, opened 2025-01-09
removed fill from line icons 

andreitr:
Screenshot 2025-01-09 at 2 15 16 PM


src/icons/CheckCircleBrokenIcon.tsx
```diff
@@ -22,6 +22,7 @@ export const CheckCircleBrokenIcon = ({
         stroke-width="2"
         stroke-linecap="round"
         stroke-linejoin="round"
+        fill="none"
```
mcgingras, line 25: if we set fill to none then is there any point in passing fill in through as a prop?

mcgingras: or would fill={fill} make more sense maybe? Not sure if that would work without firing it up

andreitr: @mcgingras these icons are stoke-only. I am using strong={fill} as a general color settings.

mcgingras (approved):

PR #718, opened 2025-01-13
Updated RPC, Bundler, Paymaster paths and chain defs 

andreitr:
Pulled the latest version of Derive and updated all of Agora’s paths to match it. Note: still need to confirm that the conduit key is no longer needed.

### DIFF
--- a/src/lib/tenant/configs/ui/derive.ts
+++ b/src/lib/tenant/configs/ui/derive.ts
@@ -56,7 +56,7 @@ export const deriveTenantUIConfig = new TenantUI({
 
   smartAccountConfig: {
     bundlerUrl: isProd
-      ? "https://bundler-lyra-mainnet-0.t.conduit.xyz"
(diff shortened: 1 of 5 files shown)


src/lib/tenant/configs/ui/derive.ts
```diff
@@ -55,7 +55,10 @@ export const deriveTenantUIConfig = new TenantUI({
   ],
 
   smartAccountConfig: {
-    bundlerUrl: "https://bundler-prod-testnet-0eakp60405.t.conduit.xyz",
+    bundlerUrl: isProd
+      ? `https://bundler-lyra-mainnet-0.t.conduit.xyz/${process.env.NEXT_PUBLIC_CONDUIT_KEY}`
+      : "https://bundler-prod-testnet-0eakp60405.t.conduit.xyz",
+
     entryPointAddress: "0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789",
     factoryAddress: "0x000000893A26168158fbeaDD9335Be5bC96592E2",
```
mcgingras, line 63: do these stay the same for prod vs dev too?

PR #735, opened 2025-01-15
Fixes PTC on admin

mcgingras:
## Description
The original issue was "[setting proposal type in the UI hangs](https://voteagora.atlassian.net/browse/ENG-685)" --> the issue was that trying to update values on a given proposal type via the admin panel would silently crash giving the illusion that it was hanging forever.

The issue was that the ABI for the current proposalTypesConfigurator is incorrect – it’s missing the last param for setProposalType – module. 

The fix is to update the ABI and pass the module address to `setProposalType` so it does not silently crash.

I also noticed that it fails if you are not the admin of the PTC. It's not obvious who the admin is, so I added another bit of UI to show who the admin is so you know which wallet to switch to.

### Tenants that support admin route
- boost (`0xA622279f76ddbed4f2CC986c09244262Dba8f4Ba`)
- cyber (`0xCe8fe86767fCc91DAD9195831AcE8D89ac604C2e`)
- derive (`0x648BFC4dB7e43e799a84d0f607aF0b4298F932DB`)
- op (`0xe538f6f407937ffDEe9B2704F9096c31c64e63A8`)
- pguild (`0x32B6d1CCbFB75aa0d52e036488b169597f0fE3d0`)
- scroll (`0xA622279f76ddbed4f2CC986c09244262Dba8f4Ba `)

### Steps to test

- Ensure branch is rebased (only Michael has to do this)
- Go to the [op deployment](https://agora-next-optimism-git-eng-721-voteagora.vercel.app/) (I'm targeting OP because it's the only deployment where I have access to the admin/manager)
- Navigate to /admin
- Make sure you are logged in to the correct admin address above (`0xe538f6f407937ffDEe9B2704F9096c31c64e63A8`)
- Try updating the name or number values for a proposal type. You should be able to successfully update the proposal type.

### Known issues
- [x] admin panel colors are off in derive
- [x] why does OP not have an admin? Are they on a different version of Agora governor? (Answer is yes, they are on an older governor and PTC that uses `manager` instead of `admin`

### DIFF
--- a/src/components/Admin/ProposalTypeSettings.tsx
+++ b/src/components/Admin/ProposalTypeSettings.tsx
@@ -9,10 +9,11 @@ import ProposalType from "./ProposalType";
 import { useReadContract } from "wagmi";
 import { useAccount } from "wagmi";
 import Tenant from "@/lib/tenant/tenant";
+import { TENANT_NAMESPACES } from "@/lib/constants";
 
 const RestrictedCallout = () => {
   const { address, isConnected } = useAccount();
-  const { contracts, slug } = Tenant.current();
+  const { contracts, namespace } = Tenant.current();
 
   const { data: managerAddress } = useReadContract({
     address: contracts.governor?.address as `0x${string}`,
@@ -40,7 +41,7 @@ const RestrictedCallout = () => {
   ];
 
   // OP is is v 0.1 of agora gov and uses a different PTC that is manager and not admin based
-  if (slug === "OP") {
+  if (namespace === TENANT_NAMESPACES.OPTIMISM) {
     addressesToRender.push({
       address: managerAddress,
       label: "Manager",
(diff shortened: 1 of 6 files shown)


mcgingras:
Colors on Derive look good as of this commit, we are rebased off of main as of this commit. 
Still unable to test the actual functionality without having access to the admin accounts.

src/components/Admin/ProposalTypeSettings.tsx
```diff
@@ -6,6 +6,75 @@ import { Plus } from "lucide-react";
...
+
+  let addressesToRender = [
+    {
+      address: timelockAddress,
+      label: "Timelock",
+    },
+  ];
+
+  // OP is is v 0.1 of agora gov and uses a different PTC that is manager and not admin based
+  if (slug === "OP") {
```
andreitr, line 43: tiny nit, can we please use `TENANT_NAMESPACES.OPTIMISM` instead of the slug 🙏

andreitr (changes requested):
Looks good! I have one small request. Moving forward, I am transitioning all slug usage to database queries only. Let’s use tenant namespace constants for UI flags instead.

src/components/Admin/ProposalTypeSettings.tsx
mcgingras: Yes, good callout!

PR #736, opened 2025-01-15
Andrei/votes lazy load

andreitr:
All vote and non-vote data requests have been moved to the client. The UI progressively enhances as data is received. All queries are cached to prevent duplicate requests. The votes `cache` is shared across charts, the votes list is cached under a different key. 

I also updated the following 
- treemap chart uses tenant styles 
- linechart uses treemap tenant 
- removed style from the ENSName component to a part component

Screenshot 2025-01-15 at 4 04 49 PM

### DIFF
--- a/src/components/Proposals/ProposalPage/OPProposalPage/ProposalVotesBar/ProposalVotesBar.tsx
+++ b/src/components/Proposals/ProposalPage/OPProposalPage/ProposalVotesBar/ProposalVotesBar.tsx
@@ -1,9 +1,10 @@
 import { Proposal } from "@/app/api/common/proposals/proposal";
 import { Vote } from "@/app/api/common/votes/vote";
+import { Bar } from "recharts";
 
 interface Props {
   proposal: Proposal;
-  votes: Vote[];
+  votes: Vote[] | undefined;
 }
 
 type Support = "FOR" | "ABSTAIN" | "AGAINST";
@@ -15,6 +16,10 @@ export default function ProposalVotesBar({ proposal, votes }: Props) {
     ABSTAIN: [],
     AGAINST: [],
   };
+
+  if (!votes) {
+    return <BarSkeleton />;
+  }
 
   votes.forEach((vote) => {
     voteCounts[vote.support as Support].push(vote);
@@ -73,3 +78,11 @@ export default function ProposalVotesBar({ proposal, votes }: Props) {
     </div>
   );
 }
+
+const BarSkeleton = () => {
+  return (
+    <div className="flex animate-pulse">
+      <div className="w-full h-[10px] bg-tertiary/10"></div>
+    </div>
+  );
+};
--- a/src/components/Proposals/ProposalPage/OPProposalPage/ProposalVotesBar/ProposalVotesBar.tsx
+++ b/src/components/Proposals/ProposalPage/OPProposalPage/ProposalVotesBar/ProposalVotesBar.tsx
@@ -59,7 +59,7 @@ export default function ProposalVotesBar({ proposal, votes }: Props) {
                     <div
                       key={`${support}-${idx}`} // use a combination of support and idx as a unique key
                       style={{ flex: `${vote.weight} 1 0%` }}
-                      className={`min-w-[1px] ${support === "FOR" ? "bg-positive" : support === "AGAINST" ? "bg-negative" : "bg-neutral"}`}
+                      className={`min-w-[1px] ${support === "FOR" ? "bg-positive" : support === "AGAINST" ? "bg-negative" : "bg-tertiary"}`}
                     ></div>
                   ))}
                 </div>
--- a/src/components/Proposals/ProposalPage/OPProposalPage/ProposalVotesBar/ProposalVotesBar.tsx
+++ b/src/components/Proposals/ProposalPage/OPProposalPage/ProposalVotesBar/ProposalVotesBar.tsx
@@ -38,7 +38,7 @@ export default function ProposalVotesBar({ proposal, votes }: Props) {
                 >
                   <div
                     style={{ flex: `1 1 0%` }}
-                    className={`min-w-[1px] ${support === "FOR" ? "bg-[#41b579]" : support === "AGAINST" ? "bg-[#db5664]" : "bg-[#666666]"}`}
+                    className={`min-w-[1px] ${support === "FOR" ? "bg-positive" : support === "AGAINST" ? "bg-negative" : "bg-tertiary"}`}
                   ></div>
                 </div>
               ))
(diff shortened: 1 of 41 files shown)


src/components/Proposals/ProposalPage/OPProposalPage/ProposalVotesBar/ProposalVotesBar.tsx
```diff
@@ -38,7 +38,7 @@ export default function ProposalVotesBar({ proposal, votes }: Props) {
                 >
                   <div
                     style={{ flex: `1 1 0%` }}
-                    className={`min-w-[1px] ${support === "FOR" ? "bg-[#41b579]" : support === "AGAINST" ? "bg-[#db5664]" : "bg-[#666666]"}`}
+                    className={`min-w-[1px] ${support === "FOR" ? "bg-positive" : support === "AGAINST" ? "bg-negative" : "bg-tertiary"}`}
```
mcgingras, line 41: nice thanks for fixing these too!

src/components/Proposals/ProposalPage/OPProposalPage/ProposalVotesSummary/ProposalVotesSummary.tsx
```diff
@@ -42,7 +50,10 @@ export default function ProposalVotesSummary({ proposal, votes }: Props) {
                 AGAINST <TokenAmountDisplay amount={results.against} />
               </div>
             </div>
-            <ProposalVotesBar proposal={proposal} votes={votes} />
+            {hasVotes && (
+              <ProposalVotesBar proposal={proposal} votes={fetchedVotes.data} />
+            )}
```
mcgingras, line 55: I think we might want a loading state here -- if you are on an approval vote page and click the votes tab the query has to run to fetch the data and it takes a few seconds. without a loading state it just looks like it's empty

src/components/Proposals/ProposalPage/OPProposalPage/ProposalVotesSummaryDetails/ProposalVotesSummaryDetails.tsx
```diff
@@ -81,7 +81,8 @@ export default function ProposalVotesSummaryDetails({
 
   return (
     <div className="flex flex-col font-inter font-semibold text-xs w-full max-w-[317px] sm:min-w-[317px] bg-wash">
-      <ProposalVotesBar proposal={proposal} votes={votes} />
+      {votes && <ProposalVotesBar proposal={proposal} votes={votes} />}
```
mcgingras, line 84: same feedback, maybe we want a loading state if there are no votes rather than hiding it

mcgingras:
What are your thoughts on preloading the react query cache on the proposal page? This way, we aren't blocking the page load, but we are also pre-warming the cache so by the time someone clicks "view chart" or "votes" the cache is already warm and there is minimal load time?

src/components/Proposals/ProposalPage/ProposalChart/ProposalChart.tsx
```diff
@@ -1,25 +1,27 @@
...
-export default function ProposalChart({
-  proposal,
-  proposalVotes,
-}: {
-  proposal: Proposal;
-  proposalVotes: PaginatedResult<Vote[]>;
-}) {
+export default function ProposalChart({ proposal }: { proposal: Proposal }) {
   const [tabIndex, setTabIndex] = useState(0);
+  const [showChart, setShowChart] = useState(proposal.status === "ACTIVE");
```
mcgingras, line 13: so this will show the chart by default is the proposal is active? Otherwise, you need to click to view the chart?

src/components/Proposals/ProposalPage/ProposalChart/ProposalChart.tsx
```diff
@@ -51,21 +53,32 @@ export default function ProposalChart({
...
+                </div>
+              )}
             </div>
-          )}
-          {tabIndex === 1 && (
-            <div className="tab-panel">
-              <TreeMapChart proposal={proposal} proposalVotes={proposalVotes} />
+          ) : (
+            <div className="text-secondary text-xs">
+              {"Loading chart data..."}
```
mcgingras, line 78: nit but we could maybe have a better loader, like a skeleton chart that pulses

src/hooks/useProposalNonVotes.ts
```diff
@@ -0,0 +1,33 @@
...
+  const { data, isFetching, isFetched } = useQuery({
+    enabled: enabled,
+    queryKey: [QK, proposalId],
+    queryFn: async () => {
+      return (await fetchVotersWhoHaveNotVotedForProposal(proposalId, {
+        offset: offset || 0,
+        limit: limit || 20,
+      })) as PaginatedResult<any[]>;
+    },
+    staleTime: 180000, // 3 minute cache
```
mcgingras, line 29: hmm does this mean if someone votes then they won't immediately see their name go away from the non-voter list? If they refresh it will wipe the cache though, so its not a huge deal? I don't think we are optimistically updating react-query cache on that mutation anyways so maybe not a big deal

src/hooks/useProposalNonVotes.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const useProposalNonVotes = ({
+  enabled,
+  limit,
+  offset,
+  proposalId,
+}: Props) => {
+  const { data, isFetching, isFetched } = useQuery({
+    enabled: enabled,
+    queryKey: [QK, proposalId],
```
mcgingras, line 22: we will want to add limit and offset here as well

mcgingras (approved):

jefag (changes requested):
I'd like to see a testing plan and the results associated with the outcome of that plan.  

I'd recommend @andreitr write down a few notes about how you tested, and suggestions for high risk things that Bryan can test.

src/components/Proposals/ProposalPage/ProposalChart/ProposalChart.tsx
andreitr: correct, this is the current functionality.

src/components/Proposals/ProposalPage/OPProposalPage/ProposalVotesSummary/ProposalVotesSummary.tsx
andreitr: I intentionally left out the loader. The idea is to implement progressive UI enhancement as the data loads. The initial view includes the data, but not the chart—the chart appears after a slight delay.

Screenshot 2025-01-16 at 9 48 37 AM

andreitr: @mcgingras i suppose I can add a pulsing placeholder bar as a loader...

mcgingras: okay thats fine then -- I think a loader for the list of voters is more important. That's what I noticed was missing when I was testing out the preview branch

src/components/Proposals/ProposalPage/OPProposalPage/ProposalVotesSummaryDetails/ProposalVotesSummaryDetails.tsx
andreitr: Loading state added

andreitr:
**QA Surface Area**
The changes are localized to the proposal page, including standard, approval, and optimistic proposals for all tenants.

The areas impacted are:
	•	All charts on the page (main and sidebar visualizations)
	•	Votes list
	•	“Hasn’t Voted” list

Screenshot 2025-01-16 at 10 53 16 AM

jefag (approved):

andreitr:
Screenshot 2025-01-16 at 2 30 40 PM

I just noticed that the new implementation (on the right) does not sort the votes by weight; instead, it displays them in the order they are received.

PR #744, opened 2025-01-20
Resolves redirect problem on OP

mcgingras:
Removes the issue of multiple state updates which was causing the page to crash. 
I believe in prod the redirect was happening, but it caught the error, then simply did not perform the page update as an alternative to showing a crashed page. While confusing, I guess it's a better experience. On local, we don't get this improvement, and the page simply crashes. This is how I was able to debug. 

I'm not 100% sure this will solve the problem, so I'm hoping deploying to a preview environment via this branch will help us check on the issue.


andreitr (approved):
This is so strange... I wonder if it has to do something with the way nextjs hydrates `back` pages. Ship if works!

bryan-agora:
QA Approved. Confirmed that the back button returns a user to the proposals page when they are on a single proposal screen.

PR #746, opened 2025-01-23
Unistaker

andreitr:
[Demo](https://www.loom.com/share/47760ba07c5f4d79a204b93a42bc42d0?sid=06bb1a4f-16f2-4ef6-92fd-0583d0fde8d2)

Screenshot 2025-01-27 at 3 13 57 PM


src/hooks/useRewardDuration.ts
```diff
@@ -0,0 +1,21 @@
+import { useQuery } from "@tanstack/react-query";
+import Tenant from "@/lib/tenant/tenant";
+
+export const REWARD_DURATION_QK = "rewardPerToken";
```
pedrohasf, line 4: queryKey seems wrong here, as it is the same as in src/hooks/useRewardPerToken.ts. but this might have been intentional?

pedrohasf (approved):
Looks good to me, just one small comment

src/app/staking/deposits/[deposit_id]/delegate/components/EditDelegateFlow.tsx
```diff
@@ -73,7 +71,7 @@ export const EditDelegateFlow = ({
       )}
 
       {step === 2 && (
-        <HStack className="grid grid-cols-1  sm:grid-cols-4 gap-5 sm:gap-10">
+        <div className="grid grid-cols-1 sm:grid-cols-4 gap-5 sm:gap-10">
```
Comment (mcgingras) [inline src/app/staking/deposits/[deposit_id]/delegate/components/EditDelegateFlow.tsx:74]: I think `HStack` is supposed to be a div with `flex flex-col` but that would also clash with `grid`? Maybe the original `HStack` is just broken 🤷

src/app/staking/[addressOrENSName]/deposits/DepositList.tsx
```diff
@@ -1,33 +1,26 @@
-"use client";
+"use server";
```
Comment (mcgingras) [inline src/app/staking/[addressOrENSName]/deposits/DepositList.tsx:1]: I don't think you need this

src/app/staking/[addressOrENSName]/deposits/DepositListAction.tsx
```diff
@@ -1,15 +1,17 @@
...
 import React from "react";
 import { useTokenBalance } from "@/hooks/useTokenBalance";
 
-interface DepositListActionProps {
+interface Props {
   address: string;
 }
 
-export const DepositListAction = ({ address }: DepositListActionProps) => {
+export const DepositListAction = ({ address }: Props) => {
```
Comment (mcgingras) [inline src/app/staking/[addressOrENSName]/deposits/DepositListAction.tsx:14]: very pedantic comment but I like to inline the props, especially if theres only a single prop. This way as a dev you can read the component, read the props, read the types, without having to then find that interface to understand the props. It's not hard to do but I find that micro-frictions add up. Super pedantic comment though, you don't have to fix

src/app/staking/[addressOrENSName]/page.jsx
```diff
@@ -1,8 +1,6 @@
 import React from "react";
```
Comment (mcgingras) [inline src/app/staking/[addressOrENSName]/page.jsx:1]: would it be possible to make this a `tsx` file?

src/app/staking/components/StakingStats.tsx
```diff
@@ -1,52 +1,53 @@
...
+import { useTotalSupply } from "@/hooks/useTotalSupply";
+import { useTotalStaked } from "@/hooks/useTotalStaked";
+import { useRewardPerToken } from "@/hooks/useRewardPerToken";
+import { useRewardDuration } from "@/hooks/useRewardDuration";
+
+export const StakingStats = () => {
+  const { data: totalSupply } = useTotalSupply({ enabled: true });
+  const { data: totalStaked } = useTotalStaked({ enabled: true });
+  const { data: rewardPerToken } = useRewardPerToken({ enabled: true });
+  const { data: rewardDuration } = useRewardDuration({ enabled: true });
```
mcgingras, line 14: could probably use `isLoading` here as well if we didn't want to default to 0

mcgingras (approved):

src/app/staking/deposits/[deposit_id]/delegate/components/EditDelegateFlow.tsx
Comment (andreitr) [reply src/app/staking/deposits/[deposit_id]/delegate/components/EditDelegateFlow.tsx:74]: Good catch. I don't think it worked as intended when I originally implemented this :)

PR #749, opened 2025-01-24
Added the Approval Options back to the past votes

kent:
Alright, this is a strange one. This was reported by a Discord user. 

CleanShot 2025-01-24 at 00 35 00@2x

On the approval vote, the options you voted for didn't show up. This used to work, but I can't find it in Git blame (probably there but eyes are hurting). 

So, the prolem was really simple, the `params` data for an approval vote is HEX encoded, and we were trying to parse the JSON directly, so it was going to `null` when passed to the client. 

This PR fixes this. 

CleanShot 2025-01-24 at 01 18 16@2x

### DIFF
--- a/src/lib/voteUtils.ts
+++ b/src/lib/voteUtils.ts
@@ -16,6 +16,7 @@ import { VotingPowerData } from "@/app/api/common/voting-power/votingPower";
 import Tenant from "@/lib/tenant/tenant";
 import { TENANT_NAMESPACES } from "@/lib/constants";
 import { Block } from "ethers";
+import { AbiCoder } from "ethers";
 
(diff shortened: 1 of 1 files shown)


kent:
Known issue: when you are logged in, your vote options don't show up. I've got the problem identified but want to get this PR out first, as it's boocking a larger delegate.

mcgingras:
The interesting thing is that this file hasn't been meaningfully touched since July. I don't remember this feature being broken for that long. Is it possible that we perhaps changed the way we are indexing such that params are now hex encoded when they previously were not? I just want to make sure if that is the case, there are not other places that might be broken.

PR #754, opened 2025-01-26
Refactored sidebar chart to use proposal values and not votes

andreitr:
Changelog

- Refactored sidebar chart to use proposal data instead of the votes (way faster) 
- Updated the votes list to load only the first 10 results (a bit faster) 
- Created a new charts endpoint that returns all voting data without any transformations
- Updated the main charts to use a separate endpoint (fixes data issues) 
- Moved chart requests to the `/api/proposals/[proposalId]` for client-side requests
- Moved data transformation to the client (block time calculation etc)
- Updated proposal status to use tenant colors 

Screenshot 2025-01-29 at 12 11 01 PM


src/components/Proposals/ProposalPage/Charts/TimelineChart.tsx
```diff
@@ -5,45 +5,204 @@ import {
...
+   * A future fix will read each tenant and stack depending on how the tenant counts quorum.
+   */
+  if (namespace === TENANT_NAMESPACES.ENS) {
+    stackIds.against = "2";
+  } else if (namespace === TENANT_NAMESPACES.UNISWAP) {
+    stackIds.against = "2";
+    stackIds.abstain = "3";
+  }
+
+  useEffect(() => {
```
pedrohasf, line 70: I think it would be cleaner to instead of using a useEffect + useState combo using a useMemo here. Would achieve the same thing but in a slightly cleaner/less complex way. Not very important

pedrohasf (approved):
Looks good to me, just one small comment

src/components/Proposals/ProposalPage/Charts/TimelineChart.tsx
andreitr: @pedrohasf the useEffect waits for the block and then performs the transformation. Are you suggesting I wrap the whole thing in `useMemo`?

pedrohasf: The suggestion would be to remove the useEffect and the useState for chartData and do the computation of the chartData value (transformation) inside of a useMemo instead of the useEffect. Something like `const chartData = useMemo(() => {` This would also allow for waiting the block and then performing the transformation, same as the useEffect

andreitr:
IMPACT: LOW RISK 🟢
The changes are isolated to the proposal page charts and do not impact critical functionalities such as voting, delegation, etc. 

QA [Optimism](https://agora-next-optimism-avd66dzdu-voteagora.vercel.app) 
- [Standard](https://agora-next-optimism-avd66dzdu-voteagora.vercel.app/proposals/61099545161324963999210479653546830615027626331636601117087361441562082554363) 
- [Approval](https://agora-next-optimism-avd66dzdu-voteagora.vercel.app/proposals/100185742509226493864823544103005937543123293361993183572321106627817282245206)
- [Optimistic - no display data](https://agora-next-optimism-avd66dzdu-voteagora.vercel.app/proposals/89833428552921894524951159614223980956121507297618142977710022171557412680990) 

QA ENS
- [Standard](https://agora-next-5c9nzgdq2-voteagora.vercel.app/proposals/100928477039597534530410979592341706732910523273914947444866837124858037213431)

QA Uniswap 
- [Standard](https://agora-next-uniswap-3at0r3qur-voteagora.vercel.app/proposals/23)
- [Standard](https://agora-next-uniswap-3at0r3qur-voteagora.vercel.app/proposals/14)

PR #756, opened 2025-01-27
Updates stale voting states loading states (ENG-1022)

mcgingras:
## Description 
Resolves the issue of some voter participation numbers number showing up, even after minutes of waiting. The issue was twofold -- the `voterStats` query was returning no record if the voter had never voted yet, which meant that the voter stats data fetch returned 0 for the field `num_proposals`. `num_proposals` was intended to be the number of proposals submitted to the tenant, so even if the voter had never voted, this should still return the correct number. The reason is that the UI was then looking at this response, and if `num_proposals` was 0 it was skipping over the render and showing nothing. This was to prevent div/by/zero. 

The fix here is two-fold
1. fix the query to return correct num_proposals even if the voter has never voted
2. ~~return 0 instead of nothing if num_proposals truly is zero.~~ Following the JIRA thread, I see Brennan mention that if there are no proposals in the tenant, we should hide voter stats. So I'm removing this "fix"

## Test cases
- [ ] open the preview branch and prod and compare numbers to make sure the query did not alter the response
- [ ] check a few delegate profiles to make sure the responses are fine there as well, and are matching the numbers shown on the voter page.
- [ ] make sure a tenant with no proposals (do we have any of these?) hides voter stats rather than showing 0%.

### DIFF
--- a/src/app/api/common/delegates/getDelegates.ts
+++ b/src/app/api/common/delegates/getDelegates.ts
@@ -580,19 +580,18 @@ async function getVoterStats(
         AND cancelled_block IS NULL
     )
     SELECT
-        v.voter,
-        sum(COALESCE((
+        COALESCE(v.voter, $1) as voter,
+        COALESCE(sum((
             SELECT count(DISTINCT last_10_props.proposal_id) AS count
             FROM last_10_props
             WHERE last_10_props.proposal_id = v.proposal_id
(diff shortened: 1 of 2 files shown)


jefag:
>(do we have any of these?)
Derive, Boost, PGuild

andreitr (approved):
LGTM

andreitr (approved):

PR #757, opened 2025-01-27
Updates to proposal votes panel UI

pedrohasf:


src/components/Proposals/ProposalPage/OPProposalPage/ProposalVotesCard/OptimisticProposalVotesCard.tsx
```diff
@@ -50,7 +50,7 @@ const OptimisticProposalVotesCard = ({
           </HStack>
         </button>
         <div>
-          <div className="py-0 px-4 font-semibold mb-2">Proposal votes</div>
+          <div className="py-0 px-4 font-semibold mb-2">Voting Activity</div>
```
mcgingras, line 53: super small nit but I typically like to only upper the first word unless it's a proper title. So in this case, `Voting activity`

src/components/shared/TokenAmountDisplay.tsx
```diff
@@ -1,22 +1,48 @@
...
+      specialFormatting
+    );
+  }, [amount, decimals, maximumSignificantDigits, specialFormatting]);
 
-  return <span>{`${formattedNumber} ${currency}`}</span>;
+  return (
+    <span
+      className={cn(
+        useChivoMono ? chivoMono.variable : "",
+        hideCurrency && icon ? "flex items-center gap-1" : ""
```
mcgingras, line 41: should this be `!hideCurrency`?
The way I'm reading this is like "If hideCurrency and icon is given, then add classes". But seems like it should be "If don't hide currency and icon is given, then show classes" 

Maybe could even be simplified to "if an icon is given, show classes". Maybe we don't need the hideCurrency

mcgingras: oh nevermind, maybe I'm misunderstanding. It seems like it either takes a currency, or an icon? So if you hide currency, then you want to show the icon?

src/components/Votes/ApprovalProposalVotesList/ApprovalProposalSingleVote.tsx
```diff
@@ -81,8 +94,24 @@ export default function ApprovalProposalSingleVote({ vote }: { vote: Vote }) {
                 </>
               )}
             </div>
-            <div className={"font-semibold text-secondary"}>
-              <TokenAmountDisplay amount={weight} />
+            <div className={"font-semibold text-primary"}>
```
mcgingras, line 97: Maybe something to get design help on but two things here:

1. I don't think the rightmost token amount looks like it's hoverable or that it has a tooltip to show. The only reason I knew to hover was because I'm reviewing this code. Maybe we need another design pass to make the hover more obvious.
2. I don't like how hovering the tooltip however the voter card at the same time. We should get rid of the voter card entirely, there was a conversation around this earlier last week. Instead, we would make the voter hoverable and link out to the delegate page. cc @andreitr on that one

Screenshot 2025-01-28 at 9 50 27 AM

src/components/Votes/ProposalVotesList/ProposalSingleVote.tsx
```diff
@@ -88,7 +111,32 @@ export function ProposalSingleVote({
                 )}
               </HStack>
               <HStack alignItems="items-center">
-                <TokenAmountDisplay amount={vote.weight} />
+                <TooltipProvider>
```
mcgingras, line 114: Same feedback about how the hover property of this token amount is not really obvious.

mcgingras: FWIW it's not your fault, I think we just need better design on this.

src/components/shared/TokenAmountDisplay.tsx
```diff
@@ -1,22 +1,48 @@
...
+  decimals?: number;
+  currency?: string;
+  maximumSignificantDigits?: number;
+  useChivoMono?: boolean;
+  hideCurrency?: boolean;
+  icon?: React.ReactNode;
+  specialFormatting?: boolean;
+};
+
 export default function TokenAmountDisplay({
```
andreitr, line 18: This component is starting to feel overloaded; it is trying to do too many things. Let’s split it into two components: maybe `TokenAmount` and `TokenAmountDecorated`.

Also, let’s avoid using the font name in the props. What if we move this into tenant customization? Perhaps a custom font or something similar?

andreitr (changes requested):
The code looks good. I made a minor architectural change request. Let me know if it makes sense. Also, please add testing notes 🙏

src/components/shared/TokenAmountDisplay.tsx
pedrohasf: Yes, definitely got a bit confusing/wrong in this code section. We didn't actually needed the hideCurrency, i want to add the classes if there is an icon and that's it, currency is unrelated. Updated it, thanks for the catch!

src/components/Proposals/ProposalPage/OPProposalApprovalPage/OptionResultsPanel/OptionResultsPanel.tsx
```diff
@@ -151,15 +153,20 @@ function SingleOption({
...
+          />
+          <span className={cn("ml-1 text-tertiary", chivoMono.variable)}>
             {percentage === 0n
-              ? "(0%)"
-              : "(" + Math.round(Number(percentage) / 100).toString() + "%)"}
+              ? "0%"
+              : (Number(percentage) / 100).toFixed(2) + "%"}
           </span>
         </div>
       </HStack>
```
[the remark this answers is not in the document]
pedrohasf: Sounds good, updated this specific file and will keep an eye open for other ones i touch to update it as we go

src/components/shared/TokenAmountDisplay.tsx
pedrohasf: Agreed, component was trying to do too many things and would be better splitted. Splitted it as recommended in TokenAmount which is only responsible for displaying the content and not the styles, and TokenAmountDecorated which is responsible for styling it.

Removed the font name in props and switched to using className.

I agree, moving it into tenant customization would be pretty neat, makes it super easy to change that font if needed for whatever reason. Updated it and now the font is coming from the tenant UI customization, super easy to update if we want to

pedrohasf:
Tested on:

Optimism standard proposal: https://agora-next-optimism-e1blcbibf-voteagora.vercel.app/proposals/35692458870234994087797547148639646668926724279854193659568886275069128375329

Optimism approval proposal: https://agora-next-optimism-e1blcbibf-voteagora.vercel.app/proposals/103233058869801302468598399790244977438277111501442249656921040560964072388879

Optimism optimistic proposal: https://agora-next-optimism-e1blcbibf-voteagora.vercel.app/proposals/18791625511772160981078190941721381202777441127220591356677323690302803867739

ENS standard proposal: https://agora-next-2lxqrrixk-voteagora.vercel.app/proposals/11953787303123140055012479667272984441675581576035225657361346001679430351375

Uniswap standard proposal: https://agora-next-uniswap-a51zom7vy-voteagora.vercel.app/proposals/23

andreitr (approved):

src/app/staking/[addressOrENSName]/deposits/DepositListAction.tsx
```diff
@@ -40,10 +40,10 @@ export const DepositListAction = ({ address }: DepositListActionProps) => {
...
                   <span className="mr-2">
-                    <TokenAmountDisplay amount={depositedBalance} /> staked
+                    <TokenAmountDecorated amount={depositedBalance} /> staked
                   </span>
                 )}
-                <TokenAmountDisplay
+                <TokenAmountDecorated
                   maximumSignificantDigits={4}
                   amount={tokenBalance || 0n}
                   currency={""}
```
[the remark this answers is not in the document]
Comment (pedrohasf) [reply src/app/staking/[addressOrENSName]/deposits/DepositListAction.tsx]: Nice catch, thank you! We should use hideCurrency in this case, changed it

pedrohasf:
QA approved and ready to merge - intentionally holding for today according to the plan here https://voteagora.slack.com/archives/C08952X5YD6/p1738700529173339

PR #758, opened 2025-01-27
UX update to expanding chart

pedrohasf:
Just a small low priority UX update of something that was bothering me. Not sure if this is the behavior we want but it looks a lot better to me.

Before:
https://share.cleanshot.com/htccjvkr

After:
https://share.cleanshot.com/YlqYHc92

### DIFF
--- a/src/components/Proposals/ProposalPage/ProposalChart/ProposalChart.tsx
+++ b/src/components/Proposals/ProposalPage/ProposalChart/ProposalChart.tsx
@@ -27,26 +27,38 @@ export default function ProposalChart({ proposal }: { proposal: Proposal }) {
     setShowChart((prevState) => !prevState);
   };
 
+  const handleButtonClick = (e: React.MouseEvent<HTMLButtonElement>) => {
+    if (showChart) {
+      e.stopPropagation();
+    }
+  };
+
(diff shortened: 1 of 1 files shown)


jefag:
Very Nice!

That was bothering me too so much!

mcgingras (approved):
Nice! I love the "ball is in your court" attitude of shipping this PR.
One feedback I noticed in the original video is that double clicking was selecting the tab text. We might be able to get rid of it with the `select-none` utility class from tailwind. It's not totally necessary to put it in here, but I wanted to mention it anyways.

mcgingras:
The only build failure is from the `jira-linker` action, so I think it's fine if we merge this despite the "unsuccessful check"

andreitr (approved):
LOVE IT 🚢 🚢 🚢

pedrohasf:
> Nice! I love the "ball is in your court" attitude of shipping this PR. One feedback I noticed in the original video is that double clicking was selecting the tab text. We might be able to get rid of it with the `select-none` utility class from tailwind. It's not totally necessary to put it in here, but I wanted to mention it anyways.

Nice, thank you for the pointer! Agreed and updated with it

PR #761, opened 2025-01-28
Adds dev badge

mcgingras:
Adds badge to header when in a dev environment. Jeff made the suggestion somewhere in Slack. The idea is that it's not always clear if we are on prod or dev, especially when on a preview branch. This badge helps us know for certain which environment we are on.

Screenshot 2025-01-28 at 10 27 21 AM


jefag:
Can it say "test-contracts-mode" (not `dev-mode`)

People should be able to tell they are on a preview link, not production easily.  And a preview link could point at dev or prod DB.  So I don't want to overload what `dev` means.

jefag:
We need a second one, imho, for the DB.  And then, we're going to need a third one, for the JSON-RPC forks we're planning.

andreitr (approved):
love it

PR #766, opened 2025-01-29
Eng 1040 add support for bubble chart and better map view

kent:
Added support for update map view and bubble view

407593054-9827c76c-d199-4c18-aac6-330ac9d2b8b3

407592984-078d631e-44f6-4d8f-a92d-54da53e6baff


kent:
**Tested on Scroll**

CleanShot 2025-01-28 at 22 34 58@2x
CleanShot 2025-01-28 at 22 34 48@2x

kent:
**Tested on Uniswap**

CleanShot 2025-01-28 at 22 36 20@2x
CleanShot 2025-01-28 at 22 36 13@2x
CleanShot 2025-01-28 at 22 36 05@2x

kent:
Tested on ENS

CleanShot 2025-01-28 at 22 32 36@2x
CleanShot 2025-01-28 at 22 32 26@2x

kent:
Tested on OP

CleanShot 2025-01-28 at 22 48 42@2x
CleanShot 2025-01-28 at 22 47 57@2x

bryan-agora:
Testing w/ preview branches. Findings:

- No adverse impacts that I found
- Loading times for the Timeline and Map visualizations are near instant, but the Bubble chart is slow >5 seconds and the loading message feels strange "Reticulating splines..." was this intentional? [Loom](https://www.loom.com/share/6cdae8199ef94720ae2517d1b11d5347?sid=089ca545-4bcd-4443-8d08-3241aaadf1e4
)

Two other notes that weren't covered from the Figma: Expand by default and/or completely remove the expand/collapse functionality. My thought here is at least expand by default.
Some of the bubbles are incredibly small (I know it's voting power related).

kent:
> Testing w/ preview branches. Findings:
> 
> * No adverse impacts that I found
> * Loading times for the Timeline and Map visualizations are near instant, but the Bubble chart is slow >5 seconds and the loading message feels strange "Reticulating splines..." was this intentional? [Loom](https://www.loom.com/share/6cdae8199ef94720ae2517d1b11d5347?sid=089ca545-4bcd-4443-8d08-3241aaadf1e4)
> 
> Two other notes that weren't covered from the Figma:
> 
> * Expand by default and/or completely remove the expand/collapse functionality. My thought here is at least expand by default.
> * Some of the bubbles are incredibly small (I know it's voting power related).

Thanks for the feedback Bryan. Let me see if I can optimize the Bubble chart. Maybe I can have them pop in as they are loading vs. waiting for all of them to load. 

I don't think there is much we can do about the really small ones, as in OP the relative voting power is so different, but I will play with the scaling factors there too.

src/components/Proposals/ProposalPage/BubbleChart/BubbleChart.tsx
```diff
@@ -0,0 +1,292 @@
...
+import { rgbStringToHex } from "@/app/lib/utils/color";
+import { Plus, Minus, RotateCcw } from "lucide-react";
+import ENSName from "@/components/shared/ENSName";
+
+//Bubble contransts
+const CHART_DIMENSIONS = {
+  width: 720,
+  height: 230,
+  padding: 3,
+  maxVotes: 50,
```
andreitr, line 16: I'd bump this to at least 250. There is no reason to limit this since all data is already there.

src/components/Proposals/ProposalPage/BubbleChart/BubbleChart.tsx
```diff
@@ -0,0 +1,292 @@
...
+  support: "0" | "1" | "2";
+  value: number;
+  r: number;
+}
+
+const SCALING_EXPONENT = 0.4;
+const transformVotesToBubbleData = (votes: ChartVote[]): BubbleNode[] => {
+  const sortedVotes = votes
+    .slice()
+    .sort((a, b) => Number(b.weight) - Number(a.weight))
```
andreitr, line 38: Do we need the weight sorting here?

andreitr (approved):
LFG! I added a couple of non-blocking comments. Since we’re loading all votes, there’s no reason to artificially limit the data. Also, the bubble chart feels a bit squished—could we increase its size or add a full-screen button? Additionally, how about making the bubble chart the default view? It’s more informative than the timeline.

src/components/Proposals/ProposalPage/BubbleChart/BubbleChart.tsx
kent: @andreitr since there can be sooo many votes it makes the loading and graph almost unusable, so I am sorting by weight and then taking the top for there.
You thinking I could do that in the query?

andreitr: @kent, the updated query is optimized for the chart—it loads all votes and sorts them by block number. I’d be curious to see how this works without any filtering—pretty sure D3 can handle a couple thousand data points fairly easily.

PR #768, opened 2025-01-29
Fix cancelled button not clickable

pedrohasf:
Tested on https://agora-next-optimism-ylst9502a-voteagora.vercel.app/proposals/88081957117197857131618413056069851464239155680037310590555441907734514873482

The fix implemented here was to basically just remove the bottom section of the card from the hoverTrigger allowing the cancelled button to be clickable. Another option would be to display the modal in the bottom instead of on top of the card.

Chose first option due to lower risk/impact but can switch to option 2 if we decide that it's better.


mcgingras (approved):
lgtm -- long term I think we should get rid of the hovers. It's a feature that I've always found to be quite clunky.

PR #769, opened 2025-01-30
Use proposals_v3 for optimism

pedrohasf:
DO NOT MERGE

### DIFF
--- a/src/app/api/analytics/vote/getProposalVoteCounts.ts
+++ b/src/app/api/analytics/vote/getProposalVoteCounts.ts
@@ -28,7 +28,7 @@ async function getProposalVoteCounts() {
                 relevant_proposals AS
                 (
                       SELECT *
-                      FROM   uniswap.proposals_v2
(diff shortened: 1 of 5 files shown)


src/app/api/analytics/vote/getProposalVoteCounts.ts
```diff
@@ -28,7 +28,7 @@ async function getProposalVoteCounts() {
                 relevant_proposals AS
                 (
                       SELECT *
-                      FROM   uniswap.proposals_v2
```
mcgingras, line 31: I don't think we need to change this for UNI

jefag: Correct.

jefag: Although...since we're not merging this branch, it doesn't matter at all.

jefag:
Testing

PR #770, opened 2025-01-30
Fix errors with etherfi getVoterStats in the getDelegates endpoint

CarlGocht:
I was poking around vercel and noticed high rates for etherfi. The delegates endpoint has been failing with: 

ERROR: invalid input syntax for type integer: ""

https://vercel.com/voteagora/agora-next-etherfi/logs?selectedLogId=8rw5r-1738242692108-05fb0c2ea531

This seems to be due to handling of empty strings from the db, however when looking at the db proposals_v2 is either not there or an empty material view. This is because etherfi is snapshot so we aren't indexing their proposals but the error is a bit strange given that context. Maybe I am not looking at the right table/view.

### DIFF
--- a/src/app/api/common/delegates/getDelegates.ts
+++ b/src/app/api/common/delegates/getDelegates.ts
@@ -568,7 +568,7 @@ async function getVoterStats(
         SELECT proposal_id
         FROM ${namespace}.proposals_v2
         WHERE contract = $2
-        AND end_block::INTEGER <= $3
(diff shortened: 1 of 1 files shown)


jefag (changes requested):
Thanks for trying to be pro-active here, but please do not merge.  This is high-risk, low-priority to fix.

andreitr (approved):
Good catch. I will leave this to @jefag to asses the risk.

andreitr (commented):
Good catch!

PR #773, opened 2025-01-30
Addresses "back button navigation" issue

mcgingras:
tbd


andreitr (approved):
LGTM

p.s. I started replacing fetchAllForVoting with individual hooks - there is just too many things happing in that one call.

mcgingras:
One note is that voting power took a while ( > 15 mins, maybe even 30 minutes, I stepped away from my computer) to resolve. It eventually came through fine, but something might be up with the indexer. It has nothing to do with this branch, I switched branches, tested main, still faced same issue.

jefag:
Agree, it has nothing to do with this branch.

PR #779, opened 2025-02-03
Fix: Update Optimism faq link 

sudheerDev:
* Update Optimism FAQ link to `agoraxyz` instance of notion
* Add partial and advanced delegation keys in tenants of scroll, derive and OP
* Use these links in respective delegation components

Testing notes: Tested on scroll preview below with delegation


mcgingras (changes requested):
Check out `src/lib/tenant/tenantUI.ts`

```typescript
public link(name: string): UILink | undefined {
    if (this._linksCache[name] !== undefined) {
      return this._linksCache[name];
    }

    const result = this._links?.find((t) => t.name === name);
    this._linksCache[name] = result;
    return result;
  }
```

The tenant offers a way to read links, so we should be able to make this:
`const faq = ui.link("faq")`

We shouldn't have to export anything then. 
Thanks!

sudheerDev:
Aah, thank you. Will update the pr

sudheerDev:
@mcgingras Here is the link that was hardcoded was same for all tenants and it was OP Faq doc, if we have to use `ui.link('faq')` I assume we have to add the same link for all configs?

mcgingras:
Great work so far! 

The context that I would not expect you to grasp on your first day is that only a few tenants have support for partial and advanced delegation. So, I don't think all tenants need to have the optimism FAQ link. Only tenants that support advanced or partial delegation will need this link.

Here is my suggestion:

If this link is only used for explaining the partial + advanced delegation, maybe we should rename it away from `faqLink` to something like `advancedDelegationFaqLink`. This way, it becomes more clear that other tenants that do not support advanced delegation do not need this link in their config.

Next, I would check which tenants support partial delegation. If we check the `DelegateActions` component we can see that Derive and Scroll support partial delegation, and only OP supports advanced delegation.

So, I think we should make two new links... `advancedDelegationFaqLink` will be used in the advanced delegation dialogs, and `partialDelegationFaqLink` will be used for partial delegation dialogs.

I would add these links to any tenants that support each feature. Derive and Scroll would get partial and OP would get advanced.

I think we will need to create a new notion page explaining the features though -- it doesn't make sense to direct folks from Derive and Scroll to the OP notion explainer.

CarlGocht:
Partial and advanced are the same link? Just double checking to make sure thats correct

sudheerDev:
> Partial and advanced are the same link? Just double checking to make sure thats correct

Yes, it has been same before as well, also this is same for few supported tenants as well. Need to have different ones sometime

CarlGocht:
This needs some fixing still since deployments are failing

sudheerDev:
@CarlGocht Should be good now.

PR #780, opened 2025-02-03
Use default getQuorum method

jefag:


mcgingras:
Addressed by Pedros theme branch -- closing

PR #785, opened 2025-02-04
Fix: Update relativeVotingPowerToVotableSupply calculation

sudheerDev:
BigInt / BigInt in Js is truncated to nearest integer causing it to return zero. Adding a new util function calculateBigIntRatio to calculate relativeVotingPowerToVotableSupply
Added a flag in swagger to indicate `votingPowerRelativeToVotableSupply` is deprecated and returning the value in string format from a new key `relativeVotingPowerToVotableSupply`

Testing Notes:
Checked on https://agora-next-optimism-git-fix-eng-1078-votingpow-9253bf-voteagora.vercel.app/ with test wallet and can see relative vote power now
 
Untitled

### DIFF
--- a/src/app/api/common/delegates/getDelegates.ts
+++ b/src/app/api/common/delegates/getDelegates.ts
@@ -521,11 +521,12 @@ async function getDelegate(addressOrENSName: string): Promise<Delegate> {
         )
       : cachedNumOfDelegators;
 
-  const votingPowerRelativeToVotableSupplyExponential = votableSupply && BigInt(votableSupply) > 0n
-    ? Number(Number(totalVotingPower) / Number(votableSupply)).toExponential(3)
-    : 0;
-  
-  const votingPowerRelativeToVotableSupply = Number(votingPowerRelativeToVotableSupplyExponential);
+  const votingPowerRatio =
+    votableSupply && BigInt(votableSupply) > 0n
+      ? (Number(totalVotingPower) / Number(votableSupply)).toExponential(3)
+      : 0;
+
+  const votingPowerRelativeToVotableSupply = Number(votingPowerRatio);
 
   // Build out delegate JSON response
   return {
--- a/src/app/api/common/delegates/getDelegates.ts
+++ b/src/app/api/common/delegates/getDelegates.ts
@@ -521,6 +521,12 @@ async function getDelegate(addressOrENSName: string): Promise<Delegate> {
         )
       : cachedNumOfDelegators;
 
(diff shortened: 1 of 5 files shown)


sudheerDev:
@jefag I fixed the issue with relative votepower but haven't added a new key for it, since it has always been 0 before. i can add a new key if you feel strongly about it. Do let me know.

src/app/api/common/delegates/getDelegates.ts
```diff
@@ -521,6 +521,13 @@ async function getDelegate(addressOrENSName: string): Promise<Delegate> {
         )
       : cachedNumOfDelegators;
 
+  const votingPowerRatio =
+    votableSupply && BigInt(votableSupply) > 0n
+      ? (Number(totalVotingPower) / Number(votableSupply)).toExponential(3)
+      : 0;
```
stepandel, line 527: `Number(totalVotingPower)` will overflow in most cases and `Number(votableSupply)` will definitely overflow, and result in incorrect calc here.
For most of the ERC20 tokens, the number will precede by 18 digits representing decimal points. [MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) is under 16 digits

stepandel (changes requested):
Let's aim to have this number returned as `string` (even if it's very small and has a bunch of 0s)

src/app/api/common/delegates/getDelegates.ts
sudheerDev: Aah, yes. This wont work. Let me scale the number so i can divide as BigInts

sudheerDev:
@stepandel Up for review again

src/app/api/common/utils/bigIntRatio.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    const result = calculateBigIntRatio(verySmall, veryLarge);
+    expect(result).toBe("0.000000000049");
+  });
+
+  test("handles 18-digit BigInt numbers", () => {
+    const num1 = 10n;
+    const num2 = 123456789012345678n;
+
+    const result1 = calculateBigIntRatio(num1, num2);
+    expect(result1).toBe("0.000000000000000081");
```
jefag: # 🔥

jefag:
Verified new result is ~6% w/ @sudheerDev by inspecting response on his local.  Will confirm on prod too.

PR #786, opened 2025-02-04
Fix voters page filters overflow + change view refreshing query

pedrohasf:
- Slightly reduced gap between filters + slightly reduced filters button font size + slightly reduced filters width size.
- Filters components refactored to use a common component to reduce tailwind maintenance
- Changing view doesn't trigger a new query

Test notes:
- Tested on https://agora-next-uniswap-git-pedro-eng-1094-uniswap-1c274e-voteagora.vercel.app/delegates

ENG-1094 + ENG-1093

### DIFF
--- a/src/app/delegates/components/DelegatorFilter.tsx
+++ b/src/app/delegates/components/DelegatorFilter.tsx
@@ -1,12 +1,10 @@
 "use client";
 
 import { useRouter, useSearchParams } from "next/navigation";
-import { Listbox } from "@headlessui/react";
-import { Fragment } from "react";
-import { ChevronDown } from "lucide-react";
 import { useAddSearchParam, useDeleteSearchParam } from "@/hooks";
 import { useAgoraContext } from "@/contexts/AgoraContext";
 import { useAccount } from "wagmi";
+import FilterListbox from "@/components/common/FilterListbox";
 
 const FILTER_PARAM = "delegatorFilter";
 const DEFAULT_FILTER = "all_delegates";
@@ -20,7 +18,7 @@ export default function DelegatorFilter() {
   const { address } = useAccount();
 
   const filterParam = searchParams?.get(FILTER_PARAM) || DEFAULT_FILTER;
-  const delegateeFilterOptions: { value: string; sort: string }[] = [
+  const delegateeFilterOptions = [
     {
       value: "All delegates",
       sort: "all_delegates",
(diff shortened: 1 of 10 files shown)


CarlGocht:
After initially pressing the sort button the text disappears from the button for me. 

image

pedrohasf:
> After initially pressing the sort button the text disappears from the button for me.
> 
> image

Nice catch, thank you! Should be fixed now

src/components/common/FilterListbox.tsx
```diff
@@ -0,0 +1,77 @@
...
+    }
+    return (
+      options[value]?.value ||
+      Object.values(options).find((opt) => opt.sort === value)?.value
+    );
+  };
+
+  return (
+    <Listbox as="div" value={value} onChange={onChange}>
+      {() => (
```
sudheerDev, line 41: Nit: Don't think this is needed here, if you update the PR for any other changes, consider removing this. Though its fine for now, it can break memoisation if we pass any arrow functions later into these.

src/app/delegates/components/DelegatorFilter.tsx
```diff
@@ -19,8 +17,10 @@ export default function DelegatorFilter() {
   const { setIsDelegatesFiltering } = useAgoraContext();
   const { address } = useAccount();
 
-  const filterParam = searchParams?.get(FILTER_PARAM) || DEFAULT_FILTER;
-  const delegateeFilterOptions: { value: string; sort: string }[] = [
+  const currentParam = searchParams?.get(FILTER_PARAM);
+  const filterParam = currentParam ? "my_delegates" : "all_delegates";
+
+  const delegateeFilterOptions = [
```
sudheerDev, line 23: Not in the scope of this PR: lets move these outside of components next time when we add so we don't break any memoisation of child components. Here in this case it is fine since the child component is anyway not memoized, but its better practice

sudheerDev (approved):

CarlGocht:
I am getting sorting issues now. Most delegators does not return the right sorted list, and neither does most power. I checked against another dev preview from another issue and both those appear to sort correctly in other previews and on prod. 

image

image

pedrohasf:
> I am getting sorting issues now. Most delegators does not return the right sorted list, and neither does most power. I checked against another dev preview from another issue and both those appear to sort correctly in other previews and on prod.

Yes, thanks for the catch. It was a side effect of the refactoring, while for other filters we use the same value for the object key and the sort property in the sort component that wasn't the case. When i tried to make it all standard it broke due to this being different from the other filters.

Did some more proper testing this time and hopefully should be all good now

CarlGocht (approved):

PR #794, opened 2025-02-06
Fix: Round of Delegate info card participation percentage

sudheerDev:
* Round of to 2 decimal points, this would reflect in inactive and active delegate cards
 
 Testing notes: Tested locally. Once xai is merged we can use preview to test this on preview


jefag:
@sudheerDev can you rebase ontop of the XAI PR and then merge into that "feature branch"?

PR #805, opened 2025-02-10
Fix to get advanced delegations from alligator contract

CarlGocht:
Tested against optimism as the ticket calls that out specifically.

Also manually tested against cyber as a FULL delegation tenant and derive as a PARTIAL delegation tenant

### DIFF
--- a/src/app/api/common/delegations/getDelegations.ts
+++ b/src/app/api/common/delegations/getDelegations.ts
@@ -62,7 +62,7 @@ async function getCurrentDelegateesForAddress({
       findAdvancedDelegatee({
         namespace,
         address,
-        contract: contracts.token.address,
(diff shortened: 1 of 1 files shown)


CarlGocht:
One interesting note is that 2/3 of the transactions in the jira ticket exist in the database. This one does not https://optimistic.etherscan.io/tx/0x4736a97c92372a9a8485b892d4746d4abaf91e4b5efe558ac06bd666427d96e1

It is the delegation from ferrodri.eth to chappy.eth. This delegation still appears though because of another transaction a day later https://optimistic.etherscan.io/tx/0xd34ab91178a063783fc36c0d4aeca7d22eaadfa67033e6e8d020fe521a147244

I am curious if that is expected to be missing from the advanced_delegations table or not

pedrohasf (approved):

PR #808, opened 2025-02-11
Share my vote

pedrohasf:


bryan-agora:
Tested the Share my Vote functionality. It's all working as expected. The designs match the Figma, etc. Just one piece of feedback, but it's primarily related to the timing of the Share my Vote Modal pop-up and the time the vote gets submitted and updated in the UI.

I shared my vote immediately after submitting it and the image that was shared was blank (didn't include my vote). So if anyone shares their vote before their vote is recorded on chain and displayed in the UI their vote won't be included in the screenshot. 

@pedrohasf - any thoughts that would help here?

pedrohasf:
> Tested the Share my Vote functionality. It's all working as expected. The designs match the Figma, etc. Just one piece of feedback, but it's primarily related to the timing of the Share my Vote Modal pop-up and the time the vote gets submitted and updated in the UI.
> 
> I shared my vote immediately after submitting it and the image that was shared was blank (didn't include my vote). So if anyone shares their vote before their vote is recorded on chain and displayed in the UI their vote won't be included in the screenshot.
> 
> @pedrohasf - any thoughts that would help here?

Thanks for the review here!

This is one that went under my radar, i think i have an straightforward solution for it. For vote reason + support in this case i am getting the state on the UI to show it on the image + twitter/warpcast share box but the proposal results in the image are all being calculated with data from the DB which can take a few minutes to update with the new casted vote. Should be easy enough to just make sure to include the casted vote VP in the proposal results calculations when sharing. Will implement the change

src/lib/voteUtils.ts
```diff
@@ -276,3 +268,268 @@ export function getVpToDisplay(
...
+}: {
+  proposal: Proposal;
+  votableSupply?: string;
+  newVote?: Pick<Vote, "support" | "reason" | "weight" | "params">;
+}) {
+  const { token } = Tenant.current();
+
+  const formattedVotableSupply = votableSupply
+    ? Number(BigInt(votableSupply) / BigInt(10 ** token.decimals))
+    : Number(115000000);
```
sudheerDev, line 485: Any reason to hard code here?

src/hooks/useVotableSupply.ts
```diff
@@ -0,0 +1,22 @@
...
+export const useVotableSupply = ({ enabled }: Props) => {
+  const { data, isFetching, isFetched } = useQuery({
+    enabled: enabled,
+    queryKey: [QK],
+    queryFn: async () => {
+      const response = await fetch(`/api/v1/votable_supply`);
+      const data = await response.json();
+      return data.votable_supply;
+    },
+    staleTime: 1000 * 60 * 60, // 1 hour
```
sudheerDev, line 18: I made a new util for this and it is merged but my cache is for 1 min probably that can be passed in as arg and use that?

sudheerDev (approved):
Looks good to me, just a minor change req

src/lib/voteUtils.ts
pedrohasf: Not really, it's just a decent fallback using a approximate value for OP but i doubt it would ever reach that fallback. This variable is used just for optimistic proposals. I could fallback to null and if not defined use the same calculations as standard proposal but that seems a little worse. I only expect votableSupply to be false if it's a approval proposal where we don't need it and don't send it as a variable.

src/hooks/useVotableSupply.ts
pedrohasf: Leaving this here, there is a new duplicate api route apparently that also needs to be cleaned up. Let's do it out of this PR.

sudheerDev (approved):

PR #810, opened 2025-02-11
ENG-1138 Add delegate statement html sanitization

CarlGocht:
This does both client and server side sanitization and leaves urls available for use

Tested against ENS thoroughly using the given eval statement. Also tested on uniswap the same way.
Tested on ENS using the markdown that we are meant to support from another users delegate statement to make sure this does not interfere with markdown. That user is here https://agora-next-kjr7yzj80-voteagora.vercel.app/delegates/0x0fcd2ead9943194eaa8bc6702a97555aad50455b

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "@tailwindcss/typography": "^0.5.10",
     "@typechain/ethers-v6": "^0.5.1",
     "@types/d3": "^7.4.3",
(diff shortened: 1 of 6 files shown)


pedrohasf (approved):

jefag:
https://voteagora.atlassian.net/browse/ENG-1138

sudheerDev (approved):

PR #816, opened 2025-02-12
Feat: Update checkbox component

sudheerDev:
* Change checkbox component to include checkmark
  * Update delegate code of conduct to use checkbox from tabs
  * Update notification selector component to use checkbox from tabs
  
  Tested on uniswap [preview](https://agora-next-uniswap-git-sudheer-eng-1144-fix-ho-78b612-voteagora.vercel.app/delegates/edit)
  
  
Untitled


src/components/DelegateStatement/DelegateStatementBoolSelector.tsx
```diff
@@ -22,33 +23,24 @@ export default function DelegateStatementBoolSelector({
...
+            <Checkbox
+              checked={agreeCodeConduct}
+              onCheckedChange={(checked) =>
+                handleAgreeCodeConduct(checked === true ? true : false)
+              }
+              className="mr-2"
+            />
+            Yes, I agree with the {codeOfConductLink.title}
+          </label>
+        </>
```
pedrohasf, line 42: Looks like we removed the actual link to the code of conduct. Is this on purpose?

pedrohasf (commented):
Just one minor comment that i think requires a change.

Not very important but would be great if we could abstract `src/components/DelegateStatement/DelegateStatementBoolSelector.tsx` and `src/components/DelegateStatement/NotificationSelector.tsx` into a single component. Whenever we want to make changes to the UI of them in the future we will have to duplicate code to do that. There is also going to be a new checkbox in a new ticket so it will actually be 3x.

src/components/DelegateStatement/DelegateStatementBoolSelector.tsx
sudheerDev: Both are hyper linked not just the title.

pedrohasf: I don't think i understood this.

Previously we had \<a href={codeOfConductLink.url} target="_blank">
            {codeOfConductLink.title}
          \</a> but now this link no longer exists. It seems we are not using codeOfConduct.url anywhere. How can the user access this url? Is it still linked to it somewhere? I don't see it

sudheerDev: Aah, yes. I misunderstood. we need to have the <a>

src/components/DelegateStatement/DelegateStatementBoolSelector.tsx
```diff
@@ -22,33 +23,26 @@ export default function DelegateStatementBoolSelector({
...
+                >
+                  {codeOfConductLink.title}
+                </a>
+              </>
+            }
+            title={`Agree with the ${codeOfConductLink.title}`}
+            checked={agreeCodeConduct}
+            onChange={handleAgreeCodeConduct}
+          />
+        </>
```
pedrohasf, line 44: Changes look great, thank you!

Just one last minor nit: I think the link should be in the title (marker 1), not in the label (marker 2).

CleanShot 2025-02-18 at 9  10 26@2x

In the image IMO it should show in the 1 marker. But right now it's on the 2 marker. It's not a good UX for the user to click on the label expecting to select it and get redirected to a link.

sudheerDev: Sure, I wasn't sure either way tbh. I have seen hyperlinking in Terms of conditions in checkbox as well. Ideally it should have been a hyperlink making it obvious.

pedrohasf (approved):

PR #817, opened 2025-02-12
Fix: Delete proposal type

sudheerDev:
* Send empty strings inplace of name and description to delete a proposal type
  * Add onDelete and onSuccessSetProposalType to handle client vs onchain state.
  * use isClient flag as a way to remove the proposal type from UI
  * On successful delete, use isSuccess flag to remove the proposal type from state so that ui reflects the change
  
 Testing notes:
  Tested on B3 [preview](https://agora-next-b3-git-sudheer-eng-1128-fix-the-del-8b0748-voteagora.vercel.app/admin) with admin wallet.
  
 Scenarios:

1. Click on `Add another proposal type`
2. Click on  delete button without saving.

Expected: Ui should delete the newly added row without any creating a transaction.

1. Click on delete of `proposal type 5`
2. Confirm the transaction.

Expected: On success of transaction, the ui should remove the row from the ui.

1. Add a new proposal type.
2. Repeat steps of deletion

Expected: On success of transaction, the ui should remove the row from the ui.

### DIFF
--- a/src/components/Admin/ProposalType.tsx
+++ b/src/components/Admin/ProposalType.tsx
@@ -136,7 +136,7 @@ export default function ProposalType({
     isLoadingDeleteProposalTypeTransaction ||
     isLoadingSetProposalType ||
     isLoadingSetProposalTypeTransaction;
-  const isDisabled = isLoading || name == "Optimistic";
+  const isDisabled = isLoading;
 
   function onSubmit(values: z.infer<typeof proposalTypeSchema>) {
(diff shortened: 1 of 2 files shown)


pedrohasf (approved):
Code LGTM, this would need some heavy QA before merging i think

jefag:
Snuck in with this patch : https://github.com/voteagora/agora-next/pull/839

I'll merge shortly, unless Sudheer pushes something.

PR #823, opened 2025-02-12
Fix direct delegatee date for arbitrum chain

pedrohasf:


src/app/api/common/delegations/getDelegations.ts
```diff
@@ -72,22 +71,13 @@ async function getCurrentDelegateesForAddress({
...
-        );
-      }
       return {
         from: advancedDelegatee.from,
         to: advancedDelegatee.to,
         allowance: advancedDelegatee.delegated_amount.toFixed(0),
         percentage: advancedDelegatee.delegated_share.toString(),
         timestamp: latestBlock
-          ? getHumanBlockTime(blockNumber, latestBlock, true)
+          ? getHumanBlockTime(advancedDelegatee.block_number, latestBlock, true)
```
jefag, line 80: Something is wrong here, we should be referencing `advancedDelegatee` for anything related to XAI.  That's an OP concept.

jefag: Oh, I see.  We accidentally had it wrong before.  I get it.

pedrohasf: Yes, this was reverted to how it was before. The initial change shouldn't have touched it but did. This is reverting the change referencing `advancedDelegatee`. It probably didn't affect it all though because of the chain check.

jefag (approved):

CarlGocht (approved):

PR #827, opened 2025-02-13
Fix for scroll admin page quorum percentage

CarlGocht:


src/components/Admin/ProposalType.tsx
```diff
@@ -50,6 +51,23 @@ export default function ProposalType({
...
+
+  useEffect(() => {
+    async function fetchTotalSupply() {
+      if (namespace === TENANT_NAMESPACES.SCROLL && contracts.token.isERC20()) {
+        const supply = await contracts.token.contract.totalSupply();
+        setTotalSupply(supply);
+      }
+    }
+    fetchTotalSupply();
+  }, [namespace, contracts.token.contract]);
```
pedrohasf, line 64: We should probably use src/hooks/useTotalSupply.ts here instead.

CarlGocht: thanks. done

src/components/Admin/ProposalType.tsx
```diff
@@ -50,6 +51,23 @@ export default function ProposalType({
...
+
+  const formattedSupply = Number(
+    (namespace === TENANT_NAMESPACES.SCROLL
+      ? totalSupply
+      : BigInt(votableSupply)) / BigInt(10 ** 18)
+  );
```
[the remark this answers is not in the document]
CarlGocht: good call. updated for that

src/components/Admin/ProposalType.tsx
```diff
@@ -56,6 +58,16 @@ export default function ProposalType({
   onSuccessSetProposalType,
 }: Props) {
   const { namespace, contracts, token } = Tenant.current();
+  const totalSupply = useTotalSupply({ enabled: true });
```
pedrohasf, line 61: Maybe we should use a conditional here instead of hardcoded true? 

`enabled: namespace === TENANT_NAMESPACES.SCROLL`

pedrohasf (approved):
Code LGTM, just one question but non blocking

src/components/Admin/ProposalType.tsx
CarlGocht: updated that!

PR #833, opened 2025-02-16
Separate web2 and web3 prisma clients

CarlGocht:
Four new .env variables everyone who runs agora-next locally will need

```
# prod
READ_WRITE_WEB2_DATABASE_URL_PROD=
READ_ONLY_WEB3_DATABASE_URL_PROD=
# dev
READ_WRITE_WEB2_DATABASE_URL_DEV=
READ_ONLY_WEB3_DATABASE_URL_DEV=
```


src/app/lib/prisma.ts
```diff
@@ -2,13 +2,22 @@ import { PrismaClient } from "@prisma/client";
...
+  var prismaWeb3Client: PrismaClient;
 }
 
-let prisma: PrismaClient;
+const isProd = process.env.NEXT_PUBLIC_AGORA_ENV === "prod";
+const envSuffix = isProd ? "PROD" : "DEV";
+
+const readWriteWeb2Url =
+  process.env[`READ_WRITE_WEB2_DATABASE_URL_${envSuffix}`];
+const readOnlyWeb3Url = process.env[`READ_ONLY_WEB3_DATABASE_URL_${envSuffix}`];
```
jefag, line 14: So this part here, is doubling down on a pattern that we need to clarify and make more explicit, rather than abstract.

NEXT_PUBLIC_AGORA_ENV is currently wildly overloaded, and its primary purpose, is to toggle the tenant between testnet contracts and mainnet contracts.  We need to maintain the flexibility to toggle both of these, independently of test and main contracts.  Eg. I want to run test contracts from the prod DB.  Or the main contracts from the dev DB.  

I think the easiest thing to do from here, is to change `DATABASE_URL` to `dev` and `prod`, rather than an actual URL.  Or what would be even stronger, would be to say: `DATABASE_URL = dev | prod | <custom, explicit, one-off DB string>`, so that if its set to dev or prod, it toggles the `READ_WRITE_WEB*` vars like you have, but if it's set to any other string, it actually overrides the `READ_WRITE_WEB*` vars completely.  Just trying to minimize clicks in vercel.  Need strong docs on this.  But Dev Experience wise, a dev will likely populate all four READ_WRITE_WEB* on there local, and then toggle quickly with `DATABASE_URL`.

CarlGocht: I understand the idea there but one issue is that in the future you won't be able to use DATABASE_URL as a global url string. Since the dbs will actually be different and serve different purposes. At that point there would be no valid explicit-one-off-url-string that we could use in that scenario. Curious on your thoughts there.

I will update it to at least allow for dev/prod switch from DATABASE_URL variable though.

jefag: You got it correct; but here are the two scenarios I want to enable it for a) test DBs (think CI/CD or local dev) that actually have both datasets and then also b) Simpler local development, toggling multiple env vars is easier to screw up (eg change one and not the other).

jefag: Yes! You got it correct.  I was just explaining the scenario I wanted to support.  And you got it, in the implementation.  Apologies, I should have "approved" this PR.

jefag (approved):

PR #839, opened 2025-02-18
Undisables Optimistic Proposals

jefag:
This PR is a minor patch, to reveal the functionality required to delete Optimistic Proposals


jefag:
Tested by visually inspecting that the Optimistic type shows up, but not disabled.

PR #840, opened 2025-02-19
Carl/eng 1179 add api monitoring

CarlGocht:


src/lib/monitoringService.ts
```diff
@@ -0,0 +1,100 @@
...
+  labels?: Record<string, string>;
+}
+
+class MonitoringService {
+  private enabled: boolean;
+  private namespace: string;
+  private apiKey!: string;
+
+  constructor() {
+    this.enabled = process.env.ENABLE_METRICS === "true";
```
jefag, line 16: Can we change `ENABLE_METRICS` to `ENABLE_DD_METRICS`?

src/lib/monitoringService.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    const tags = Object.entries(options.labels || {}).map(
+      ([key, value]) => `${key}:${value}`
+    );
+
+    // Add global tags
+    tags.push(
+      `env:${process.env.VERCEL_ENV === "production" ? "production" : "development"}`,
+      "service:agora-next"
+    );
```
jefag, line 43: Would this be better or worse?  Does this var get named randomly for preview branches or something else I don't get?
```
    tags.push(
      `env:${process.env.VERCEL_ENV}`,
      "service:agora-next"
    );
 ```

jefag (approved):
Looks good, but minor Qs/feedback.

src/lib/monitoringService.ts
CarlGocht: That would be very similar. Your version would have 3 names, production/preview/development. Where as mine simplifies it to just two names

src/lib/monitoringService.ts
CarlGocht: Will do!

pedrohasf (approved):

PR #842, opened 2025-02-20
Fix resolve safe transaction receipt for voting

pedrohasf:
This implementation hopefully should replace all instances of waitForTransactionReceipt in the long term. Keeping it contained to voting right now as it would impact critical functions of the app and needs to be thoroughly tested


src/lib/utils.ts
```diff
@@ -469,3 +485,106 @@ export const mapArbitrumBlockToMainnetBlock = unstable_cache(
...
+  }
+
+  if (!responseJson.isSuccessful) {
+    return undefined;
+  }
+  return responseJson.transactionHash;
+};
+
+export const wrappedWaitForTransactionReceipt = async (
+  publicClient: PublicClient,
```
sudheerDev, line 570: Is there a reason why pass in client here? Since the plan is to move to use this function mostly, should we move the client to be called in this function?

pedrohasf: That's a nice improvement, updated it

sudheerDev (approved):

src/lib/utils.ts
```diff
@@ -469,3 +485,106 @@ export const mapArbitrumBlockToMainnetBlock = unstable_cache(
...
+  }>;
+};
+
+const apiNetworkName: Record<number, string> = {
+  [mainnet.id]: "mainnet",
+  [optimism.id]: "optimism",
+  [polygon.id]: "polygon",
+  [base.id]: "base",
+  [arbitrum.id]: "arbitrum",
+  [goerli.id]: "goerli",
```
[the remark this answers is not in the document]
pedrohasf: Scroll yes and added it. Derive no because it's not available in the API

PR #843, opened 2025-02-20
Anti-spam limits into gas relays

pedrohasf:
Tested with 2 wallets (one above the limit and one below) on:

https://agora-next-n0jqq9vv9-voteagora.vercel.app/proposals/67735687352329756707539052679943584106182461334310823374771499563155029489891

vs without the change in

https://agora-next-iw2b7lp9d-voteagora.vercel.app/proposals/76818066098263882979708920517360119767598381540188577320270510470572123084921

### DIFF
--- a/src/components/Dialogs/UndelegateDialog/UndelegateDialog.tsx
+++ b/src/components/Dialogs/UndelegateDialog/UndelegateDialog.tsx
@@ -20,8 +20,10 @@ import { useConnectButtonContext } from "@/contexts/ConnectButtonContext";
 import { DelegateePayload } from "@/app/api/common/delegations/delegation";
 import Tenant from "@/lib/tenant/tenant";
 import { revalidateData } from "./revalidateAction";
-import { zeroAddress } from "viem";
+import { formatEther, zeroAddress } from "viem";
 import { useSponsoredDelegation } from "@/hooks/useSponsoredDelegation";
+import { useEthBalance } from "@/hooks/useEthBalance";
+import { UIGasRelayConfig } from "@/lib/tenant/tenantUI";
 
 export function UndelegateDialog({
   delegate,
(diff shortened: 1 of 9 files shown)


src/components/Dialogs/UndelegateDialog/UndelegateDialog.tsx
```diff
@@ -48,6 +50,19 @@ export function UndelegateDialog({
...
+  const { data: sponsorBalance } = useEthBalance({
+    enabled: isGasRelayEnabled,
+    address: gasRelayConfig?.sponsorAddress,
+  });
+
+  const isGasRelayLive =
+    isGasRelayEnabled &&
+    Number(formatEther(sponsorBalance || 0n)) >=
+      Number(gasRelayConfig?.minBalance) &&
+    Number(votingPower) > Number(gasRelayConfig?.minVPToUseGasRelay);
```
sudheerDev, line 65: Woudn't the vp here zero? i assume once delegated the vp is zero. would this mean we want to make all undelegations funded by the user?

pedrohasf: We don't want that. The vp here is coming from the function call `getVotingPowerAvailableForDirectDelegationForAddress` at `src/app/api/common/voting-power/getVotingPower.ts` which is actually a balanceOf call so i wouldn't expect it to zero out but we need to properly test it.

```
async function getVotingPowerAvailableForDirectDelegationForAddress({
  address,
}: {
  address: string;
}): Promise<bigint> {
  const { contracts } = Tenant.current();
  return contracts.token.contract.balanceOf(address);
}
```

pedrohasf: Tested and looking good

sudheerDev: Aah, Got it. Thank you.

sudheerDev (approved):

PR #857, opened 2025-03-03
* Update proposal creation flow with votable supply update

sudheerDev:
* Add new contract return type for tenants to use votable supply oracle
 * Add a new component UpdateVotableSupplyOracle
 * Render the component ifn submit form and only if tenant has votable supply oracle contract
 * Return null if connected address is not the owner

### DIFF
--- a/src/app/proposals/components/UpdateVotableSupplyOracle.tsx
+++ b/src/app/proposals/components/UpdateVotableSupplyOracle.tsx
@@ -1,0 +1,108 @@
+"use client";
+
+import { useEffect, useState } from "react";
+import {
+  useAccount,
+  useReadContract,
+  useWaitForTransactionReceipt,
+  useWriteContract,
+} from "wagmi";
+
+import Tenant from "@/lib/tenant/tenant";
+import {
+  VOTABLE_SUPPLY_QK,
+  useGetVotableSupply,
+} from "@/hooks/useGetVotableSupply";
+
+import { Card, CardContent } from "@/components/ui/card";
+import { formatNumber } from "@/lib/utils";
+import { useQueryClient } from "@tanstack/react-query";
+import { UpdatedButton } from "@/components/Button";
+
+export const UpdateVotableSupplyOracle = () => {
+  const { address } = useAccount();
+  const { data: presentVotableSupply } = useGetVotableSupply();
+  const queryClient = useQueryClient();
+  const [txHash, setTxHash] = useState<`0x${string}` | undefined>(undefined);
+
+  const { contracts } = Tenant.current();
+
+  const { writeContractAsync, isPending } = useWriteContract();
+
+  const { data: votableSupplyOracle, refetch: refetchVotableSupplyOracle } =
+    useReadContract({
+      address: contracts.votableSupplyOracle?.address as `0x${string}`,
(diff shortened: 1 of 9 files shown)


src/app/proposals/components/UpdateVotableSupplyOracle.tsx
```diff
@@ -0,0 +1,108 @@
...
+      address: contracts.votableSupplyOracle?.address as `0x${string}`,
+      abi: contracts.votableSupplyOracle?.abi,
+      functionName: "votableSupply",
+    });
+
+  const { data: ownerAddress } = useReadContract({
+    address: contracts.votableSupplyOracle?.address as `0x${string}`,
+    abi: contracts.votableSupplyOracle?.abi,
+    functionName: "owner",
+  });
```
pedrohasf, line 43: Let's try to always use chainId in the useReadContract calls, don't really matter here i think but just a good practice and can avoid some weird behavior.

src/app/proposals/components/UpdateVotableSupplyOracle.tsx
```diff
@@ -0,0 +1,108 @@
...
+  // Format the votableSupplyOracle for display
+  const formattedOracleSupply =
+    votableSupplyOracle !== undefined && votableSupplyOracle !== null
+      ? formatNumber(votableSupplyOracle.toString(), 18, 4, false, false)
+      : "0";
+
+  const formattedPresentVotableSupply =
+    presentVotableSupply !== undefined && presentVotableSupply !== null
+      ? formatNumber(presentVotableSupply.toString(), 18, 4, false, false)
+      : "0";
```
pedrohasf, line 58: `token.decimals` instead of hardcoded 18

pedrohasf (commented):
Non-blocking comments

src/app/proposals/components/UpdateVotableSupplyOracle.tsx
sudheerDev: That makes sense, will add that

src/app/proposals/components/UpdateVotableSupplyOracle.tsx
sudheerDev: Good call

sudheerDev:
@pedrohasf made couple of changes along with the above ones. Since this component only renders when votable supply contract exist for tenant, passing it from parent so we don't have to do optional chaining checks. Up for review again

pedrohasf (approved):

src/app/proposals/components/UpdateVotableSupplyOracle.tsx
```diff
@@ -18,43 +17,69 @@ import { Card, CardContent } from "@/components/ui/card";
...
   const formattedPresentVotableSupply =
     presentVotableSupply !== undefined && presentVotableSupply !== null
-      ? formatNumber(presentVotableSupply.toString(), 18, 4, false, false)
+      ? formatNumber(
+          presentVotableSupply.toString(),
+          18,
+          tokenDecimal,
+          false,
+          false
+        )
```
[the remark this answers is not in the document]
sudheerDev: Aah, thank you!

PR #878, opened 2025-03-11
Proposer can cancel in BravoGov

pedrohasf:


src/app/proposals/components/BravoGovCancel.tsx
```diff
@@ -19,8 +19,10 @@ export const BravoGovCancel = ({ proposal }: Props) => {
   const { address } = useAccount();
 
   const { data: adminAddress } = useGovernorAdmin({ enabled: true });
+  const proposer = proposal.proposer;
   const canCancel =
-    adminAddress?.toString().toLowerCase() === address?.toLowerCase();
+    adminAddress?.toString().toLowerCase() === address?.toLowerCase() ||
+    proposer?.toString().toLowerCase() === address?.toLowerCase();
```
jefag, line 26: Won't this mean that the proposer of any proposal, see the cancel-button?

jefag: I don't think that's correct, we actually need to do a check agains the bravo governor.

pedrohasf: This is already checked before getting to this component. This is the BravoGovCancel component. We also have AgoraGovCancel and AgoraOptimismGovCancel.

jefag (approved):

PR #879, opened 2025-03-12
Add balance of info in delegate to row

sudheerDev:
Add token balance column to delegated to table

Tested on [xai](https://agora-next-xai-git-sudheer-eng-1282-delegate-t-78d94d-voteagora.vercel.app/delegates/0x4c8ad181a8b891de450a50e31a003d0e530a6359) and [OP](https://agora-next-optimism-git-sudheer-eng-1282-deleg-5117b7-voteagora.vercel.app/delegates/0xd477cf2d2499620a2736bc9c0e7cf6939c29f4f9) 
Screenshot 2025-03-12 at 4 28 35 PM
Screenshot 2025-03-12 at 4 28 40 PM

### DIFF
--- a/src/components/Delegates/Delegations/DelegationToRow.tsx
+++ b/src/components/Delegates/Delegations/DelegationToRow.tsx
@@ -7,26 +7,22 @@ import { format } from "date-fns";
 import Link from "next/link";
 import { ArrowTopRightOnSquareIcon } from "@heroicons/react/20/solid";
 import ENSName from "@/components/shared/ENSName";
-import { DELEGATION_MODEL } from "@/lib/constants";
 
 export default function DelegationToRow({
   delegation,
-  delegationModel,
+  tokenBalance,
 }: {
   delegation: Delegation;
-  delegationModel: string;
+  tokenBalance?: bigint;
 }) {
   return (
     <TableRow>
-      {(delegationModel === DELEGATION_MODEL.PARTIAL ||
-        delegationModel === DELEGATION_MODEL.ADVANCED) && (
-        <TableCell>
-          {TokenAmountDisplay({
-            amount: delegation.allowance,
-            maximumSignificantDigits: 3,
-          })}{" "}
-        </TableCell>
-      )}
+      <TableCell>
+        {TokenAmountDisplay({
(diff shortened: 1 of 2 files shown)


src/components/Delegates/Delegations/DelegationToRow.tsx
```diff
@@ -7,26 +7,22 @@ import { format } from "date-fns";
...
     <TableRow>
-      {(delegationModel === DELEGATION_MODEL.PARTIAL ||
-        delegationModel === DELEGATION_MODEL.ADVANCED) && (
-        <TableCell>
-          {TokenAmountDisplay({
-            amount: delegation.allowance,
-            maximumSignificantDigits: 3,
-          })}{" "}
-        </TableCell>
-      )}
```
pedrohasf, line 29: Why remove it for these delegation models? Is this not being used?

sudheerDev: We are showing for everyone now instead of few delegation models.

pedrohasf: Shouldn't it be a conditional then? It seems to me that previously it was more correct for these delegation models

sudheerDev: I am not sure i am following here. instead of showing only for few tenants the delegated value, now we are showing balanceOf for all tenants.

sudheerDev: i understand, this is what was suggested as per the ticket to add though @pedrohasf

pedrohasf: Hmmm ok, got it. Will mark this as approved then but i think we should raise the topic one last time before merging this. It doesn't seem right to me

pedrohasf (approved):

src/components/Delegates/Delegations/DelegationToRow.tsx
sudheerDev: @pedrohasf Aah, i missed this post when i merged. I did ask about this before in standup as well wwith @jefag . happy to bring it up again though

PR #881, opened 2025-03-12
Don't push transaction signature for Gov Bravo

pedrohasf:


src/lib/constants.ts
```diff
@@ -122,3 +124,5 @@ export const retroPGFSort = {
 };
 
 export const disapprovalThreshold = 12;
+
+export const GOV_BRAVO_TENANTS: TenantNamespace[] = [TENANT_NAMESPACES.UNISWAP];
```
jefag, line 128: Ideally we're not maintaining a list of hard-coded tenants, but instead this logic lives in the TenantConfig.  

One options would be with pseudo logic something like...

timelockcontract = GOV_CONTRACT.read_timelock_contract()
timelock_style = get_timelock_style(timelockcontract)

if timelock_style not in ('Bravo', 'OZ', 'ENS-whatever'):
   throw new Exception("Unsupported style: ....")

Another option would be...if we want to hardcode it per tenant on setup, then what we should do implement the same way DELEGATION_MODEL, where it's an enum set in TenantConfig.

Right?

pedrohasf: Agreed that this is better.

I had explored the first option with a similar pseudo logic but couldn't find a really good option.

Using a setup similar to DELEGATION_MODEL with an enum is better than my solution. Updated it.

src/lib/constants.ts
```diff
@@ -122,3 +124,10 @@ export const retroPGFSort = {
 };
 
 export const disapprovalThreshold = 12;
+
+export enum GOVERNOR_TYPE {
+  AGORA = "AGORA",
+  ALLIGATOR = "ALLIGATOR",
+  BRAVO = "BRAVO",
+  ENS = "ENS",
+}
```
jefag, line 133: ALLIGATOR handles delegation logic, it's toggled with  DELEGATION_MODEL, which is normally token-based logic.

I think we need what you did, but for other reasons.  

The logic we need to toggle, is actually `TIMELOCK_TYPE`.

I looked at these four timelocks...

Uniswap - https://etherscan.io/address/0x1a9C8182C09F50C8318d769245beA52c32BE35BC#code
ENS - https://etherscan.io/address/0xFe89cc7aBB2C4183683ab71653C4cdc9B02D44b7#code
Scroll - https://scrollscan.com/address/0x79D83D1518e2eAA64cdc0631df01b06e2762CC14#code
Optimism - https://optimistic.etherscan.io/address/0x0eDd4B2cCCf41453D8B5443FBB96cc577d1d06bF#code

The Enum should be (something like...)

`TIMELOCK_NO_ACCESS_CONTROL` # Uni, 
`TIMELOCKCONTROLLER_WITH_ACCESS_CONTROL` # ENS,
`TIMELOCKCONTROLLER_WITH_ACCESS_CONTROL_ERC721_ERC115` # Default (Optimism, Scroll, all other agora implementations)

jefag (changes requested):

jefag (approved):

PR #888, opened 2025-03-17
Remove dssave param

pedrohasf:
Added https://github.com/pmndrs/zustand as a dependency to manage this state.

It wasn't necessarily needed and could use React's useContext instead but opted to use Zustand instead just for ease of use  and better/more powerful in general. Useful in many other places


jefag (approved):
Nice!

PR #897, opened 2025-03-19
Feat: Updates for delegate page filtering and sorting

sudheerDev:
PR adds new sorting and filtering, All filters, such as My delegates, Endorsed delegates, issues and stakeholders are part of one filter dropdown driven by tenant configs.  Issues, stakeholders filter count is displayed in the respective collapsible headers along with the aggregated number on filter button. Also adds a cross button on delegate filter to clear entire filter state.  Mobile experience differs slightly with apply button action needed to persist filter changes. 

Key changes:

API Updates
* Added hasStatement parameter to delegate API for filtering delegates with statements

UI Components
- FilterResetBox: New component with reset functionality in the header
- DelegatesFilter: Main filter component combining issues, stakeholders, and button filters
- DelegatesSortFilter: Updated sort filter with new styling and reset capability
- CitizensSortFilter: Updated sort filter for citizens with new styling
- BadgeCount: Component to display number of selected filters on larger screens
- MobileCountIndicator: Compact count display for smaller screens
- DelegatesFilterChips: Chips for issues, stakeholders, and delegate filters with delete option
- MobileDropdownFilter: Animated bottom sheet filter using Framer Motion

Hooks
- useDelegatesFilter: Manages active filter state for issues, stakeholders, and button filters
- useDelegatesSort: Handles sorting state and logic
- Updated useAddSearchParam and useDeleteSearchParam with args to take in multiple params.

Mobile Experience
- Implemented temporary state management for mobile filters
- Users must explicitly apply or cancel filter changes. 
- All filters are part of one filters drawer, which includes sort as well for delegates. Citizens sort filter will still be part of popover experience. 
- Only one of the grid or layout icon is shown allowing user to toggle from the state they are in.    

Testing notes.
Uniswap: has issues and stakeholder as filter options to test with, but no test instance data atm but can verify with url updates.
Optimism: Citizens tab.
Scroll: endorsed filter.
Xai: for colour contrast.

Note: Left active voter filter out of scope from implementation.


src/app/api/common/delegates/getDelegates.ts
```diff
@@ -96,15 +97,31 @@ async function getDelegates({
...
-      const delegateStatementFiler =
-        filters?.endorsed || filters?.issues || filters?.stakeholders
+      // Add hasStatement filter condition
+      const hasStatementFilterQuery = filters?.hasStatement
+        ? `
+          AND s.payload IS NOT NULL
+          AND s.payload != '{}'::jsonb
+          AND s.payload ? 'delegateStatement'
+          AND s.payload ->> 'delegateStatement' != ''
+          AND s.dao_slug = '${slug}'
```
jefag, line 107: This will work, but we're redundantly adding dao_slug criteria to all of these filters.  That's redundant, and likely slower.

Take a look at the final query that gets built in different cases, you should be able to make it filter on dao_slug only once, regardless of user-options.

sudheerDev: Thank you, makes sense. I updated the code to remove slug and have it just once in the query

jefag:
RE: Testing Notes

> Uniswap: has issues and stakeholder as filter options to test with, but no test instance data atm but can verify with url updates.

You can test against the production DB for this feature, it's read-only.

> Optimism: Citizens tab.
> Scroll: endorsed filter.
> Xai: for colour contrast.

Is there an action item here?

sudheerDev:
'Optimism: Citizens tab.
Scroll: endorsed filter.
Xai: for colour contrast.'

Mentioning tenants that can be used for different filters

pedrohasf (changes requested):
I haven't finished reviewing all of the code here but i have a big request:

Can we use [nuqs](https://nuqs.47ng.com/) to handle URL state instead of how we are doing it here? We already use it in parts of the code and it would make a lot of the code on this PR easier to read and maintain. Safer, less lines of code, etc. I strongly feel like we should use it here but if you think it's too big a change then let's talk about it

src/components/Delegates/DelegatesTabs/DelegatesTabs.tsx
```diff
@@ -1,105 +1,137 @@
...
   const [tab, setTab] = useQueryState("tab", {
     defaultValue: "delegates",
     shallow: false,
     startTransition,
   });
   const [layout, setLayout] = useQueryState("layout", {
     defaultValue: "grid",
-    shallow: true,
+    shallow: false,
   });
```
pedrohasf, line 30: Why change the shallow prop here? This doesn't seem right to me

src/components/ui/dropdown-menu.tsx
pedrohasf, line 1: I think the components from shadcn here in the /ui folder should be mostly left intact.

If it's not a change that can keep the same behavior as before then IMO it should live in a separate file. Changes that change the behavior of these based on a property should be fine.

Concern is keeping these as reusable as possible

src/components/Delegates/DelegatesFilter/CitizensSortFilter.tsx
```diff
@@ -0,0 +1,82 @@
...
+    );
+    setIsOpen(false);
+  };
+
+  const resetSort = (e: React.MouseEvent) => {
+    setIsDelegatesFiltering(true);
+    router.push(deleteSearchParam({ name: "citizensOrderBy" }), {
+      scroll: false,
+    });
+  };
```
pedrohasf, line 43: Missed updating this one to nuqs, it was the only it seems

pedrohasf (changes requested):
Thanks for the changes, look great! Just a few nits now

src/components/Delegates/DelegatesTabs/DelegatesTabs.tsx
sudheerDev: Just was making it consitent. since i wanted to capture all route changes on server as well but true, this does not require a fetch. will revert this

src/app/delegates/DelegatesPageClient.tsx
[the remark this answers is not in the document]
sudheerDev: Aah!

src/components/Delegates/DelegatesFilter/CitizensSortFilter.tsx
sudheerDev: Done. thank you!

src/hooks/useAddSearchParam.ts
```diff
@@ -11,17 +11,29 @@ export const useAddSearchParam = () => {
...
       clean,
     }: {
-      name: string;
-      value: string;
+      name?: string;
+      value?: string;
+      params?: Record<string, string>;
       clean?: boolean;
     }) => {
-      const params = new URLSearchParams(
```
sudheerDev, line 20: I will revert this in the next review comment. We don't think we will use these changes from here on

src/components/ui/dropdown-menu.tsx
pedrohasf: Fair. Let's ignore this comment.

pedrohasf (approved):

PR #900, opened 2025-03-19
ENG-1387 ens info page missing proposal threshold quorum

jefag:
This PR:

- Sets a distinct "About Us" for UNI
- Fixes the proposal types table for UNI & ENS...sortof


jefag:
image

PR #902, opened 2025-03-21
Update proposal creation with wrappedWaitForTransactionReceipt hook

sudheerDev:


src/app/proposals/sponsor/components/BasicProposalAction.tsx
```diff
@@ -89,10 +96,10 @@ const BasicProposalAction = ({
           }
         }}
       >
-        Submit proposal
+        Sponsor
```
pedrohasf, line 99: Is this copy change intentional?

sudheerDev: Aah no. Fixed it. Update the similar change of enabled for other proposal types as well. Thank you!

sudheerDev:
@pedrohasf up for review again

pedrohasf (approved):

PR #909, opened 2025-03-24
* Add Canopy branding for demo instance

sudheerDev:


src/app/info/components/InfoAbout.tsx
```diff
@@ -70,7 +70,9 @@ const InfoAbout = () => {
           </div>
           <div className="sm:w-1/2">
             <h3 className="text-lg font-bold text-primary capitalize">
-              About {namespace}
+              {namespace === TENANT_NAMESPACES.DEMO
+                ? "About Canopy"
+                : "About " + namespace}
```
pedrohasf, line 75: I think using `brandName` makes more sense here. It would propagate for all tenants as well which should be fine because `brandName` has a fallback for `namespace` already

pedrohasf (approved):

src/app/info/components/InfoAbout.tsx
sudheerDev: Aah didn't knew we have brandname. made the change

PR #916, opened 2025-03-27
Set default notifications preferences to prompt

pedrohasf:


jefag (approved):
Not sure how this was missed, but seems good to me.

PR #921, opened 2025-03-28
ENG-1359: Don't include optimistic proposals when calculating voter stats

CarlGocht:
I think this really is just a one liner change. Maybe someone has cleaned up some of the complexity here already


jefag:
But it could handle 3 of 3.

PR #929, opened 2025-04-03
Add max height to modal

pedrohasf:
We should test as many modals as possible before merging this but confident that it won't break anything.


sudheerDev:
@pedrohasf is there a link where we can check? 95vh still seems quite a lot for modal. just want to see how it feels

pedrohasf:
> @pedrohasf is there a link where we can check? 95vh still seems quite a lot for modal. just want to see how it feels

Agreed that 95vh is quite a lot for a modal but i want to fix the issue with minimal impact in places i am not aware of. 95vh gets the job done. Not a link because i have to impersonate but here are a couple screenshots:

simulator_screenshot_69315B13-49F2-46D6-A9B2-92E1119CE3DC

CleanShot 2025-04-09 at 16 53 36@2x

PR #944, opened 2025-04-10
ENG-1424: Use delegators table for num delegators

CarlGocht:
Requires https://github.com/voteagora/queries/pull/169/files to be deployed before this will show accurate data


jefag:
@CarlGocht is this PR still needed?

PR #950, opened 2025-04-11
Fix width overflows in mobile proposal page

pedrohasf:
This PR fixes a few width overflows in mobile proposal page => some new, some old


pedrohasf:
Will merge this PR as is without reviews due to timing (friday night)

I have spot checked a few proposals both mobile and desktop

Super low risk as it's just styles.

PR #955, opened 2025-04-15
Undelegate button for partial delegation

pedrohasf:
Partial delegation button always said "Delegate" without considering if it was already a delegatee


jefag:
Dumb question - does this flow through to update the button on both the delegate profile page and the delegate listing page?

pedrohasf:
> Dumb question - does this flow through to update the button on both the delegate profile page and the delegate listing page?

It does

CleanShot 2025-04-16 at 14 57 12
CleanShot 2025-04-16 at 14 57 20

jefag (approved):

PR #958, opened 2025-04-16
Display create proposal button in OP for all

pedrohasf:


sudheerDev:
@pedrohasf are we sure we want to enable it for all addresses? looks like it was intentionally removed before

PR #981, opened 2025-04-29
Remove ignoring 0x00000 delgatees

jefag:
This PR reverts a prior patch, to attempt to fix undelegation.


jefag:
TODO - Need to suppress 0x0000s from the from/to-table.

PR #985, opened 2025-04-29
ENG-1571-Action-preview-truncates-to-only-function-call-and-doesn-t-show-params

Daktic:
This adds additional functionality to the action preview to show param variable names and values into the preview. If there are more than 3 parameters in the function signature, it will produce an ellipsis, but will still be seen from the tooltip.

The getFunctionSignature now returns an interface with the function name, the parameters in tuple format [name, value] if they exist, and a function to build a string representation of the functionSignature (used in the tooltip).

### DIFF
--- a/src/components/Proposals/ProposalPage/ApprovedTransactions/ProposalTransactionDisplay.tsx
+++ b/src/components/Proposals/ProposalPage/ApprovedTransactions/ProposalTransactionDisplay.tsx
@@ -585,11 +585,14 @@ const ActionSummary = ({
               <div className="truncate text-xs break-all">
                 <p>{functionSignature.functionName}(</p>
                 {functionSignature.paramValues &&
-                  functionSignature.paramValues.map( (param, idx) => (
+                  functionSignature.paramValues.slice(0, 3).map( (param, idx) => (
                   <p key={idx} className="font-medium" style={{ marginLeft: '1em' }}
-                    >{param[0]}={param[1]}</p>
+                    >{param[0]}={param[1]},</p>
                   )
                 )}
+                {functionSignature.paramValues && functionSignature.paramValues.length > 3 &&
+                  <p className="font-medium" style={{ marginLeft: '1em' }}>...</p>
+                }
                 <p>)</p>
               </div>
             </TooltipTrigger>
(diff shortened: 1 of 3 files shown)


src/lib/utils.ts
```diff
@@ -629,17 +637,52 @@ export function getFunctionSignature(decodedData: any): string | null {
...
-      ([_, param]: [string, any]) => {
-        return param.type || "unknown";
+    const paramTypes: {
+      paramValues: [string, string] | null;
+    }[] = Object.entries(decodedData.parameters).map(
+      ([paramName, paramValue]: [string, any]) => {
+        // Force null if undefined
+        if (paramName === undefined || paramValue.value === undefined) {
+          return { paramValues: null };
+        }
```
pedrohasf, line 647: It's quite common for the paramName to be undefined/null. Can we use the type if that's the case as a fallback instead of sending this as null?

pedrohasf (commented):
Code looks great! Excited to have your first PR up!

Just one small comment for an improvement

src/components/Proposals/ProposalPage/ApprovedTransactions/ProposalTransactionDisplay.tsx
pedrohasf, line 1: We also use `getFunctionSignature` in the `src/lib/seatbelt/checks/check-decode-calldata.ts` file => can you also update that to consider the new function return interface please

src/lib/utils.ts
jefag: If the `paramName` is null, another option would be to just omit it completely, and just show the value.

src/lib/utils.ts
```diff
@@ -629,17 +637,52 @@ export function getFunctionSignature(decodedData: any): string | null {
...
+        };
       }
     );
-    signature += paramTypes.join(",");
-    signature += ")";
 
-    return signature;
+    const functionName = decodedData.function.toString();
+
+    const toStringValue = () => {
```
Daktic, line 656: If there are no paramtypes, i could just use the function name directly. Then I don't have to handle it elsewhere

src/lib/utils.ts
Daktic: I have it handling each individual case based on if a name, value, or neither is supplied. if neither a name or a value are supplied it will fallback to using the param type.

I didn't see any examples of this other than having no params in the current proposal dashboard.

Ill merge this afternoon if no objections!

Daktic:
TODO: add unit tests to tuple parameters in function call

PR #989, opened 2025-05-01
garrett/1539-gov-editability

Daktic:
This change removes the ability to edit the page to change the Voting Period and Voting Delay.

The GovernSettings was given a GovernorLockedSetting component to templatize the component. 

Capitalized Headings. Open to reverting depending on site themes and standards, just a personal preference.

There is state functions in here for controlling the values for voting period and delay. I am purposely leaving that functionality in case it is used in the future. It seems to be interlaced with other in use functions so did not comment it out.

### DIFF
--- a/src/components/Admin/ProposalTypeSettings.tsx
+++ b/src/components/Admin/ProposalTypeSettings.tsx
@@ -63,7 +63,7 @@ const RestrictedCallout = () => {
     !addressesToRender.some((addr) => addr.address === address)
   ) {
     return (
-      <div className="text-sm w-[800px] rounded border-2 p-4 bg-negative/10 border-negative text-negative mt-4">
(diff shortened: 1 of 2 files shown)


src/components/Admin/ProposalTypeSettings.tsx
```diff
@@ -63,7 +63,7 @@ const RestrictedCallout = () => {
     !addressesToRender.some((addr) => addr.address === address)
   ) {
     return (
-      <div className="text-sm w-[800px] rounded border-2 p-4 bg-negative/10 border-negative text-negative mt-4">
+      <div className="text-sm max-w-[800px] rounded border-2 p-4 bg-negative/10 border-negative text-negative mt-4 overflow-x-auto overflow-y-hidden">
```
Daktic, line 66: This allows it do render better on mobile

PR #990, opened 2025-05-01
Garrett/eng 1539 gov editability

Daktic:
This change removes the ability to edit the page to change the Voting Period and Voting Delay.

The GovernSettings was given a GovernorLockedSetting component to templatize the component.

Capitalized Headings. Open to reverting depending on site themes and standards, just a personal preference.

There is state functions in here for controlling the values for voting period and delay. I am purposely leaving that functionality in case it is used in the future. It seems to be interlaced with other in use functions so did not comment it out.

### DIFF
--- a/src/components/Admin/GovernorSettings.tsx
+++ b/src/components/Admin/GovernorSettings.tsx
@@ -81,7 +81,7 @@ export default function GovernorSettings() {
     );
   }
 
-  const { data: setVotingPeriodConfig, isError: setVotingPeriodError } =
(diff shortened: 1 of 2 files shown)


pedrohasf (approved):

Daktic:
@pedrohasf I added the ability to edit that values back into the page in their own component. I also added toasts, though in the past I used an async call instead of the one used here. Everything seems to be in working order, could you just double check my work?

pedrohasf (approved):
LGTM! Thank you!

PR #992, opened 2025-05-02
Fix VP fallback

pedrohasf:
If fetching VP fails we should hide it instead of displaying 0


src/components/Proposals/ProposalPage/ApprovalCastVoteDialog/ApprovalCastVoteDialog.tsx
```diff
@@ -52,12 +52,14 @@ export function ReviewApprovalVoteDialog({
               {selectedOptions.length > 1 && "s"}
             </p>
           </div>
-          <div className="flex flex-col">
-            <span className="text-xs self-end">with</span>
-            <span className="mt-[2px]">
-              <TokenAmountDecorated amount={votingPower} />
-            </span>
-          </div>
+          {votingPower ? (
```
sudheerDev, line 55: Woudn't this cause flash on display?

sudheerDev (approved):

PR #993, opened 2025-05-02
garrett/ENG-1538-UI-to-Transfer-the-Governor-Manager-Role

Daktic:
Added a section to the admin page called Account Actions.

This will conditionally render a series of actions that can be performed on the account. If the user clicks the action button they will be prompted via a dialogue to transfer the manager address. They must input an address. Underneath is a warning that the action is reversible. They click acknowledge, and then the button changes to transfer which prompts the wallet to perform the tx. After success, it prompts a toast blockscan url, then logs out the user.

I tried to redirect the user back home but was not able to.

### DIFF
--- a/src/components/Admin/AdminAccountActions.tsx
+++ b/src/components/Admin/AdminAccountActions.tsx
@@ -1,0 +1,61 @@
+"use client";
+
+import { Button } from "@/components/ui/button";
+import { Fragment } from "react";
+
+export default function AdminAccountActions() {
+  const handleAccountTransfer = () => {};
+
+  return (
+    <section className="gl_box bg-neutral">
+      <h1 className="font-extrabold text-2xl text-primary">Account Actions</h1>
+      <p className="text-secondary">
+        Perform administrative actions on your account
+      </p>
+      <AccountActions
+        title={"Transfer"}
+        description={"Transfer account role to another address"}
+        actions={[{ title: "Transfer", onClick: handleAccountTransfer }]}
+      />
+    </section>
+  );
+}
+
+interface AccountActionButtonProps {
+  title: string;
+  onClick: () => void;
+}
+
+function AccountActionButton(props: AccountActionButtonProps) {
+  return (
+    <div className="flex ml-10">
+      <Button title={props.title} onClick={props.onClick}>
+        {props.title}
+      </Button>
+    </div>
+  );
+}
+
+interface AccountActionsProps {
+  title: string;
+  description: string;
+  actions: Array<{ title: string; onClick: () => void }>;
+}
+
+function AccountActions(props: AccountActionsProps) {
+  return (
+    <Fragment>
+      <div className="flex justify-between items-center">
+        <p className="text-sm font-semibold text-primary">{props.title}</p>
+        <p className="text-sm text-muted-foreground p-2">{props.description}</p>
+        {props.actions.map((action, index) => (
+          <AccountActionButton
+            key={index}
+            title={action.title}
+            onClick={action.onClick}
+          />
+        ))}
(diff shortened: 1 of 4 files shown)


src/components/Admin/AdminAccountActions.tsx
```diff
@@ -0,0 +1,110 @@
...
+    chainId: contracts.governor?.chain.id,
+  }) as { data: `0x${string}` };
+
+  // Can move to a util map if used elsewhere
+  switch (slug.toUpperCase()) {
+    case "OP":
+      if (address === managerAddress) {
+        accountActionToggles.transfer = true;
+      }
+  }
```
pedrohasf, line 46: This works too but the pattern we usually follow is using `namespace` from `Tenant.current()` and comparing with `TENANT_NAMESPACES` constant in the `src/lib/constants.ts` file.

pedrohasf (commented):
Just some nits

src/components/Admin/AdminAccountActions.tsx
Daktic: I like this approach better. I updated this to use a switch to return the authority for transfer for comparison based on the governor. 
`  // Compare governor_type to derive transfer authority
  const getTransferAuthority = () => {
    switch (contracts.governorType) {
      case GOVERNOR_TYPE.AGORA:
        return adminAddress;
      case GOVERNOR_TYPE.ALLIGATOR:
        return managerAddress;
      case GOVERNOR_TYPE.ENS:
      //     todo
      case GOVERNOR_TYPE.BRAVO:
      //     todo
      default:
        return adminAddress;
    }
  };

  if (address === getTransferAuthority()) {
    accountActionToggles.transfer = true;
  }`
  
  Where can I find the contracts for ENS and Bravo to check if the transfer authority is manager or admin?

src/components/Admin/AdminAccountActions.tsx
```diff
@@ -0,0 +1,110 @@
...
+    functionName: "manager",
+    chainId: contracts.governor?.chain.id,
+  }) as { data: `0x${string}` };
+
+  const { data: adminAddress } = useReadContract({
+    address: contracts.governor?.address as `0x${string}`,
+    abi: contracts.governor.abi,
+    functionName: "admin",
+    chainId: contracts.governor?.chain.id,
+  }) as { data: `0x${string}` };
```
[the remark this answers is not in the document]
Daktic: Utilized with the above impl.

src/components/Admin/AdminAccountActions.tsx
pedrohasf: Nice, i like it! Thank you!

We don't support the admin page for `GOVERNOR_TYPE.ENS` and `GOVERNOR_TYPE.BRAVO` => it would be a bigger ticket to even figure out what are the options/functions. I wouldn't worry about it.

Daktic:
### Summary:
This pull request introduces a new Account Actions section to the admin page, enabling users to safely transfer the governor manager role through an intuitive UI flow.

### Key Changes:

- Account Actions Section: A new section on the admin page that conditionally displays available actions for account management.
- Transfer Manager Role UI:
- Users can initiate the transfer of the manager role using an action button.
- A dialog prompts users to input the new manager address and displays a warning indicating that the action is irreversible.
- Upon user acknowledgment, the button updates to trigger the transaction via a wallet prompt.
- Post-Transaction Feedback:
- After a successful transaction, a toast notification displays the Blockscan URL.
- The user is logged out after the transfer is completed.

### Known Limitations:

- Currently, the user is not redirected back to the homepage post-logout due to implementation constraints.

PR #1001, opened 2025-05-06
Remove delgatees

jefag:


jefag:
This is a do-over of #995

PR #1002, opened 2025-05-06
Use lazy-loaded imports for API

jefag:
# Goal

The goal of this PR is to remove the external dependencies during the build.  Specifically, avoid/eliminate the database, and eliminate the need for JSON-RPC. This is so that we can build and deploy a test-stack, that boot from a fork.

The PR basically does 4 things:

1. Add a line `// export const dynamic = 'force-dynamic'; // this line is uncommented for e2e tests`, that gets commented out in the Dockerfile for e2e tests only.
2. Moves prisma-related imports into lazy-loaded imports.  The assumption here is that these are serverless, and these imports are dwarfed in performance time compared to establishing a DB connection.
3. Cleans up a bit of garbage related to 404 and 500s.
4. [Adds this to the next.js config](https://github.com/voteagora/agora-next/pull/1002/files#diff-882b2c04b01b2e8b2cdcf1817c30ea503a8005f1c0d54cfff37053b6801fea85R63).

...the outcome being nothing is statically generated for the e2e image, and nothing in the regular deployment changes materially.

# Development Approach

1. I did a first pass with Windsurf, asking it to do most of the work.  That was pretty much the initial commit in this PR, then after that, it was a bunch of me brute-forcing discovery of issues, using dynamic-page loading as a target state.
2. Brute force, correcting AI's mistakes + over-aggressive changes, then brute-force the rest manually.

# Asks to Reviewers

1. Will any behaviour in the e2e tests be compromised by `force-dynamic`?
2. Did I miss anything? 
3. Is anything I did redundant or unnecessary?
4. The build now works without an alchemy key.  But it fails without a DATABASE_URL, mostly because our application logic actually mandates needing it.  Is there a clever way we could improve this behaviour, without compromising next-js behaviour.


src/app/delegates/[addressOrENSName]/page.tsx
```diff
@@ -25,7 +25,7 @@ import VotesContainerWrapper, {
   VotesContainerSkeleton,
 } from "@/components/Delegates/DelegateVotes/VotesContainerWrapper";
 
-export const dynamic = "force-dynamic";
```
Comment (pedrohasf) [inline src/app/delegates/[addressOrENSName]/page.tsx:28]: Uncommented `export const dynamic = "force-dynamic";` here

src/app/delegates/edit/page.tsx
```diff
@@ -1,7 +1,7 @@
 import CurrentDelegateStatement from "@/components/DelegateStatement/CurrentDelegateStatement";
 import Tenant from "@/lib/tenant/tenant";
 
-export const dynamic = "force-dynamic";
```
pedrohasf, line 4: Uncommented `export const dynamic = "force-dynamic";`

src/app/delegates/page.jsx
```diff
@@ -5,9 +5,10 @@ import DelegateCardWrapper, {
 } from "@/components/Delegates/DelegateCardList/DelegateCardWrapper";
 import Hero from "@/components/Hero/Hero";
 
-export const dynamic = "auto"; //nuqs does not consider params changes for filters otherwise
```
pedrohasf, line 8: While the `export const dynamic` here is uncommented in the change which is good we are changing the value => i would expect some small performance degradation from this

Could we do a conditional here?

src/app/events/page.jsx
```diff
@@ -1,3 +1,6 @@
+
+export const dynamic = 'force-dynamic'
+
```
pedrohasf, line 3: I think this one should be commented?

src/app/proposals/[proposal_id]/page.tsx
```diff
@@ -14,7 +16,6 @@ import { calculateVoteMetadata } from "@/lib/voteUtils";
 import { format } from "date-fns";
 import React from "react";
 
-export const dynamic = "force-dynamic";
 export const maxDuration = 60;
```
Comment (pedrohasf) [inline src/app/proposals/[proposal_id]/page.tsx:19]: Uncommented `export const dynamic`

src/app/events/page.jsx
pedrohasf: Looks like this file should actually be deleted

src/app/proposals/create/page.jsx
```diff
@@ -1,3 +1,5 @@
+export const dynamic = 'force-dynamic';
+
```
pedrohasf, line 2: Nice, we didn't had it before but i think this is a page where it's good to have it uncommented and it applying in both app and tests. A create page should always be fully dynamic

src/app/proposals/draft/[id]/page.tsx
```diff
@@ -15,7 +17,6 @@ import { fetchProposalTypes } from "@/app/api/common/proposals/getProposals";
 import { PLMConfig } from "@/app/proposals/draft/types";
 
 export const maxDuration = 60;
-export const dynamic = "force-dynamic";
```
Comment (pedrohasf) [inline src/app/proposals/draft/[id]/page.tsx:20]: Uncommented `export const dynamic`

src/app/staking/new/page.tsx
```diff
@@ -1,4 +1,4 @@
-"use server";
+// export const dynamic = 'force-dynamic'; // this line is uncommented for e2e tests
```
pedrohasf, line 2: Not sure what's the impact of this but i am pretty sure there will be side effects here from removing `use server`. If it's blocking the build maybe we can add a new rule to the dockerfile to comment `use server` lines?

src/components/shared/ResourceNotFound/ResourceNotFound.tsx
```diff
@@ -1,3 +1,5 @@
+'use client';
+
```
pedrohasf, line 2: Small performance degradation from doing this. Probably negligible here and not worth the effort of creating a new rule for it.

src/app/staking/new/page.tsx
pedrohasf: I would say this one is not negligible and will create weird side effects. Vs the`'use client'` comment below which is negligible

src/app/admin/page.tsx
```diff
@@ -1,9 +1,10 @@
+// export const dynamic = 'force-dynamic'; // this line is uncommented for e2e tests
+
 import AdminForm from "@/components/Admin/AdminForm";
 import { fetchVotableSupply as apiFetchVotableSupply } from "@/app/api/common/votableSupply/getVotableSupply";
 import { fetchProposalTypes } from "@/app/admin/actions";
 import Tenant from "@/lib/tenant/tenant";
 
-export const dynamic = "force-dynamic";
 export const revalidate = 0;
```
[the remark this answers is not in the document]
jefag: Correct, I'll revert.  Thanks for catching.

src/app/api/analytics/metric/[metric_id]/[frequency]/route.ts
```diff
@@ -1,23 +1,22 @@
 import { type NextRequest, NextResponse } from "next/server";
-import { authenticateApiUser } from "@/app/lib/auth/serverAuth";
-import { apiFetchMetricTS } from "@/app/api/analytics/metric/[metric_id]/[frequency]/getMetricsTS";
 
-export async function GET(request: NextRequest) {
+export async function GET(
+  request: NextRequest,
+  { params }: { params: { metric_id: string; frequency: string } }
+) {
+  const { authenticateApiUser } = await import("@/app/lib/auth/serverAuth");
```
Comment (sudheerDev) [inline src/app/api/analytics/metric/[metric_id]/[frequency]/route.ts:7]: Can we make these parallel as well? just incase if there is an impact

src/app/staking/new/page.tsx
jefag: Could a solution here be a simpler, set it to `export const dynamic = 'force-dynamic';` without being commented out. That's what I've patched at the tip of this branch.

src/app/api/analytics/metric/[metric_id]/[frequency]/route.ts
Comment (jefag) [reply src/app/api/analytics/metric/[metric_id]/[frequency]/route.ts:7]: Good idea.

But - I'm going to leave that for a follow-up optimization.  I don't want to invest in optimizing this, if it doesn't fully function for some other reason.

src/app/staking/new/page.tsx
pedrohasf: Looked at it more closely and i think this file actually didn't need "use server". Having it or removing wouldn't matter actually.

I am fine with the initial change.

But since it's a "new"/create page it's also fine to have it always dynamic (what you are doing at the tip).

This file LGTM

PR #1011, opened 2025-05-08
Handle tuples in functionSignature

pedrohasf:


jefag (approved):

jefag:
@pedrohasf I think we should ship this right now.

PR #1015, opened 2025-05-12
garrett/ENG-1632-passed-proposal-should-not-be-cancellable

Daktic:
Added case for Optimistic proposal type. This could also be null but I'd rather have the feedback to indicate the proposal status.

added an edge case that will catch non-optimistic succeeded proposals. 

defaults the render label to null. used if one of the PROPOSAL_STATUS states is passed that has not been explicitly defined.


pedrohasf (commented):
In line 180 it would also be good to handle it in the same way.

It probably won't display anyway due to line 128 but makes things more explicit

PR #1017, opened 2025-05-13
Transaction simulation: cache + perf improvements + remove beta

pedrohasf:
- Removed beta tag
- Performance improvements => no specific measurements taken but visibly faster in a 5-10x (varies from proposal to proposal) order of magnitude => one specific code section was **very** slow due mostly to dumb code `(fetchStorageLayout)`
- Added cache where possible
- 	Not caching the report specifically due to high size (> 2mb). Could do it but feels like overkill => the caches added in supporting functions make it fast

### DIFF
--- a/src/lib/seatbelt/checks/check-targets-no-selfdestruct.ts
+++ b/src/lib/seatbelt/checks/check-targets-no-selfdestruct.ts
@@ -1,6 +1,8 @@
 import { bullet, toAddressLink } from "../report";
 import type { ProposalCheck } from "../types";
 import { Provider } from "ethers";
+import { getCodeCached } from "./check-targets-verified-etherscan";
+
 /**
  * Check all targets with code if they contain selfdestruct.
  */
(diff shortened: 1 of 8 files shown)


src/lib/seatbelt/checks/check-targets-no-selfdestruct.ts
```diff
@@ -43,22 +45,39 @@ async function checkNoSelfdestructs(
...
-    else if (status === "empty")
-      warn.push(bullet(`${address}: EOA (may have code later)`));
-    else if (status === "safe")
-      info.push(bullet(`${address}: Contract (looks safe)`));
-    else if (status === "delegatecall")
-      warn.push(bullet(`${address}: Contract (with DELEGATECALL)`));
-    else if (status === "trusted")
-      info.push(bullet(`${address}: Trusted contract (not checked)`));
-    else error.push(bullet(`${address}: Contract (with SELFDESTRUCT)`));
+
```
Daktic, line 59: since you are reusing bullet and res.addressLink in every path, you could use a factory and function to build the bullet.

```typescript
const createSettledResultBuilder = (addressLink: string) => {
  return (msg: string): string => {
    return bullet(`${addressLink}: ${msg}`);
  };
};
```

In the loop:
```typescript
const res = result.value;
const settledResultBuilder = createSettledResultBuilder(res.addressLink);
```

then call each path as
```typescript
info.push(settledResultBuilder("EOA"));
warn.push(settledResultBuilder("EOA (may have code later)"))
error.push(settledResultBuilder("Contract (with SELFDESTRUCT)"))
```

very nitpicky, just a suggestion

src/lib/seatbelt/checks/check-targets-verified-etherscan.ts
```diff
@@ -44,15 +53,34 @@ async function checkVerificationStatuses(
...
-      info.push(bullet(`${address}: EOA (verification not applicable)`));
-    else if (status === "verified")
-      info.push(bullet(`${address}: Contract (verified)`));
-    else info.push(bullet(`${address}: Contract (not verified)`));
+  for (const result of settledResults) {
+    if (result.status === "fulfilled") {
+      const res = result.value;
+      if (res.status === "eoa")
+        info.push(
+          bullet(`${res.addressLink}: EOA (verification not applicable)`)
```
Daktic, line 70: Same note as `check-targets-no-selfdestruct.ts`

src/lib/seatbelt/simulate.ts
```diff
@@ -3,10 +3,11 @@ import { getAddress } from "@ethersproject/address";
 import { HashZero } from "@ethersproject/constants";
 import { keccak256 } from "@ethersproject/keccak256";
 import { toUtf8Bytes } from "@ethersproject/strings";
+import { unstable_cache } from "next/cache";
```
Daktic, line 7: This file in general may be a good candidate for WASM execution if this functionality is expanded.

Daktic (approved):
Everything LGTM, I added a pattern that may make some logic in the check-target files easier to read and reuse. 

Also wondering if you can bypass a lot of the extrnal / cached calls if they are running the same sim on execution. Is it possible for the simulation to change, or is it always rendered from the same proposal?

pedrohasf:
> since you are reusing bullet and res.addressLink in every path, you could use a factory and function to build the bullet.

Tbh not a big fan of the pattern. It adds complexity and makes it less readable IMO. It's not in every path also, bullet is used in other places without the address

> This file in general may be a good candidate for WASM execution if this functionality is expanded.

Agreed, we should keep in mind as an option but feels overkill right now.

> Also wondering if you can bypass a lot of the extrnal / cached calls if they are running the same sim on execution. Is it possible for the simulation to change, or is it always rendered from the same proposal?

For existing (not new/draft) proposals it's always rendered from the same one and there are thing we could do to bypass it and make it even better like you are suggesting here. It would be only for existing proposals though and these caches that are now in place are useful for both. Even if we end up adding this "bypass" in the near future these caches will still be useful for existing proposals. Feels to me out of scope for this ticket and if needed we can revisit it later. It wouldn't mean changes to what's here but additions.

It would probably be a blob storage or similars due to the size of it. Tenderly API return is > 2 MB by itself which is huge.

PR #1022, opened 2025-05-14
Passed proposal state after queued for > 10 days

pedrohasf:
- Passed proposal state after queued for > 10 days
- Allow proposal tx simulation with no "real actions"
	- It was initially intentionally disabled because i didn't see value in having that. I changed my mind and can see the value now.

### DIFF
--- a/src/lib/proposalUtils.ts
+++ b/src/lib/proposalUtils.ts
@@ -768,7 +768,8 @@ export type ProposalStatus =
   | "PENDING"
   | "QUEUED"
   | "EXECUTED"
-  | "CLOSED";
+  | "CLOSED"
+  | "PASSED";
 
 export async function getProposalStatus(
   proposal: ProposalPayload,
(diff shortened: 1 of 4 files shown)


src/lib/proposalUtils.ts
```diff
@@ -787,11 +873,45 @@ export async function getProposalStatus(
...
+      if (queuedBlockTime) {
+        queueEventTimeSeconds = Math.floor(queuedBlockTime.getTime() / 1000);
+      }
+    }
+
+    if (
+      queueEventTimeSeconds &&
+      latestBlock.timestamp - queueEventTimeSeconds > TEN_DAYS_IN_SECONDS
+    ) {
+      if (checkHasNoCalldata()) {
```
Daktic, line 908: What makes the lack of call data equate to Passed? Because there is nothing to execute, there is no way to progress passed Queued? I believe this was discussed early but did not fully understand the problem.

Daktic (approved):
LGTM, left one question about crux of issue for my own benifit.

PR #1031, opened 2025-05-19
Etherscan v2

pedrohasf:
Etherscan v1 is being deprecated May 31st 2025


src/lib/seatbelt/encode-state.ts
```diff
@@ -400,9 +402,11 @@ async function fetchStorageLayout(
     "1": mainnet,
     "11155111": sepolia,
     "10": optimism,
-    "534352": scroll,
-    "8453": base,
+    "11155420": optimismSepolia,
     "42161": arbitrum,
+    "421614": arbitrumSepolia,
+    "8453": base,
+    "534352": scroll,
```
Daktic, line 409: If you find this to become unwieldy over time, it looks like you can reference the id values from viem/chains. would not have to support each one individually.

Daktic (approved):

PR #1034, opened 2025-05-21
Better error display on tx simulation

pedrohasf:
- Fix validation for address => wasn't working
- Do validation onBlur instead of onSubmit
- Correct validation for calldata
- Display error message on failed tx simulation toast
- Fetch token decimals when custom token instead of defaulting to 18

### DIFF
--- a/src/app/proposals/draft/components/TransferTransactionForm.tsx
+++ b/src/app/proposals/draft/components/TransferTransactionForm.tsx
@@ -23,6 +23,7 @@ import {
 } from "@/components/ui/select";
 import { Switch } from "@/components/shared/Switch";
 import { TenantToken } from "@/lib/types";
+import { getPublicClient } from "@/lib/viem";
 
 const transferABI = [
   {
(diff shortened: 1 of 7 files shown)


src/app/proposals/draft/components/TransferTransactionForm.tsx
```diff
@@ -70,13 +83,40 @@ const TransferTransactionForm = ({
...
+          });
+          setCustomTokenDecimals(Number(tokenDecimals));
+        } catch (error) {
+          console.error("Error fetching token decimals:", error);
+          setCustomTokenDecimals(18); // Fallback to 18 on error
+        }
+      };
+      fetchDecimals();
+    } else if (!useCustomToken) {
+      setCustomTokenDecimals(null); // Reset when not using custom token
```
Daktic, line 117: Is there a scenario in which useCustomToken is truthy, and customTokenAddress && isAddress(customTokenAddress) is not? 

That would only be an issue if customTokenDecimals was properly set, then somehow the second two qualifiers were changed to bad values. Likely none issue.

Daktic (approved):
checkout comment on TransferTransactionForm. otherwise everything LGTM.

src/app/proposals/draft/components/TransferTransactionForm.tsx
pedrohasf: The scenarios that come to mind:

1. Before any customTokenAddress is inputted => no issue
2. customTokenAddress was set and changed to a bad value => customTokenDecimals would retain the previous value which doesn't seem like an issue since the user will be blocked from proceeding anyway due to the zod validation. When changing back to a valid customTokenAddress customTokenDecimals would then be set to the new valid value.

PR #1038, opened 2025-05-23
Joint house checkbox

pedrohasf:


src/lib/eas.ts
```diff
@@ -0,0 +1,263 @@
...
+  tiers: number[];
+  signer: JsonRpcSigner;
+  onchain_proposalid?: string | null;
+}) {
+  const id = keccak256(
+    defaultAbiCoder.encode(
+      ["bytes32", "bytes32"],
+      [keccak256(toUtf8Bytes(slug)), keccak256(toUtf8Bytes(description))]
+    ) as `0x${string}`
+  );
```
jefag, line 66: Can we save the id as a string represented as a uint256?  Would be easier for the humans to join between these and atlas schema.  Per [here](https://voteagora.slack.com/archives/C08SXUHKYSZ/p1749088529618629?thread_ts=1749087445.831659&cid=C08SXUHKYSZ).

PR #1045, opened 2025-05-29
garrett/ENG-1484-Delegate-incorrectly-marked-as-Inactive

Daktic:
Updated the card header to use values derived from dao-node.

This fix ensures we use the correct denominator when calculating delegate statistics, preventing delegates from being incorrectly marked as inactive.

The changes:
1. Use standardized denominator values from the DAO node API
2. Fix edge cases where zero values were causing display issues

Resolves ENG-1484.


Daktic:
Had a lot of fun with merge conflicts

PR #1049, opened 2025-05-30
ENG-1587: global button standardization

aarna-arora:
## Problem

- The application had inconsistent button styles across different components, and some buttons weren't following the approved design system.
- There was a mobile layout issue with the manager wallet address in basic proposal types.

## Changes

- **Button Component Refactor:**  
  Removed the old `UpdatedButton` implementation and renamed it to `OldButton` for now.
- **Standardized Button Usage:**  
  Introduced new `UpdatedButton` (which is now the standardized button).
- **Consistent Button Styles:**  
  Ensured all buttons use the new `UpdatedButton` component and follow the approved design system (rounded, correct colors, etc.).
- **Mobile Layout Fix:**  
  Fixed a mobile layout bug where the manager wallet address wasn't wrapping properly in basic proposal types.

## Testing

- Verified consistent button styling across the application.
- Checked that all primary buttons use the new rounded design.
- Confirmed the manager wallet address now wraps properly on mobile devices.
- Ensured no regressions in existing button functionality.


jefag:
You'll need to install a tool called `prettier` to lint your code.  Can either run manually, or have VS code automatically do it for you.

sudheerDev (changes requested):
Hey @aarna-arora, Thank you!. Changes look great. 
I have a few suggestions. Updating UpdatedButton would be the Idle choice.  From what I see there are a few buttons which need to be updated still.   
Screenshot 2025-06-20 at 12 42 18 PM

That being said, after replacing all the changes, there is nothing wrong with the final code. However, one issue is that the fewer changes we make overall, the easier it is for others during major refactors. Other long-running open PRs might have conflicts since this touches quite a few files.

Once you replace the other UpdatedButton components, you can move your new design system file into UpdatedButtons and export it as UpdatedButton. This way, the changes that end up in git will be minimal.

Once you replace other Updatedbuttons, you can replace your new DS file into UpdatedButtons and export as UpdatedButton. This way the changes that end up in git will be less, if that makes sense.

If you look at this change below. 
Screenshot 2025-06-20 at 12 47 10 PM
Screenshot 2025-06-20 at 12 47 20 PM

If you follow my suggestion:

* Since primary is your default variant, you can omit it.
* Change the size to small as the default prop (which is mostly used in your changes) and omit this
* Leave fullWidth={true} for now.

The above two changes will not be part of changes anymore.  
From what I can guess, this will result in smaller changes and easier to read as well. 
Let me know if you have any questions or if this causes any issues

PR #1056, opened 2025-06-03
Fix: Duplicate toast on governor settings.

sudheerDev:
* Seperate the toasts useEffect so voting period and delay act with thier own changes.
* Issue: since the second time a transaction is created, this triggers a change in dependencies of useEffect and enters success of previous transaction block as well. causing a duplicate toast.


Daktic (approved):
LGTM

PR #1123, opened 2025-07-22
Switch to consuming everything from DAO-Node

jefag:
This branch, despite being opened on July 22, was actually branched off of `main`, early June.

We paused to add "off-chain-voting" for Op, but, basically need to get this logic into:

- the shape branch
- then main

Per this plan:

https://www.notion.so/agoraxyz/Landing-the-AG-2-0-DAO-Node-Planes-together-218528b28fc48030af8fe79df631feeb


jefag:
This PR is illustrative/communication-only.

PR #1129, opened 2025-07-24
Smooth styling between the links mimics a sliding effect.

Daktic:
Adds a transition element onto the navbar slider so that the overlay over the HeaderLink slides to the selection rather than blinks to it.


Daktic:
Short Video Demoing what it does
https://www.loom.com/share/b9bf0e58731a44878a36dd5823a158e0

PR #1162, opened 2025-08-22
Forum categories management and feature flag

pedrohasf:


src/app/admin/archived/page.tsx
```diff
@@ -5,26 +5,45 @@ import {
...
 export default async function ArchivedDataPage() {
+  const { ui } = Tenant.current();
+
   let archivedReports: ForumTopic[] = [];
   let archivedDocuments: any[] = [];
   let archivedCategories: ForumCategory[] = [];
 
+  if (!ui.toggle("forum") && !ui.toggle("duna")) {
+    return <div>Route not supported for namespace</div>;
+  }
```
Daktic, line 29: should this lead to 404 page instead?

src/hooks/useForum.ts
```diff
@@ -832,6 +836,206 @@ export const useForum = () => {
...
+        toast.success("Category created successfully!");
+        return {
+          id: result.data!.id,
+          name: result.data!.name,
+          description: result.data!.description || undefined,
+          archived: result.data!.archived,
+          adminOnlyTopics: result.data!.adminOnlyTopics,
+          createdAt: result.data!.createdAt.toISOString(),
+          updatedAt: result.data!.updatedAt.toISOString(),
+        };
```
Daktic, line 880: If there is a possibility of result data being malformed, you could catch it above this and return an error. Would be able to avoid forcing result.data!

Daktic (dismissed):

PR #1172, opened 2025-09-03
Fix: Verify Signatures with messages for verification.

sudheerDev:
* Verify signature with expected messages in delegation statement and Auth verify route.
 * Add a helper function to construct message from payload in delegation statement.


src/app/api/common/delegateStatement/createDelegateStatement.ts
```diff
@@ -25,6 +26,24 @@ export async function createDelegateStatement({
...
+  });
+
+  if (message !== expectedMessage) {
+    throw new Error("Message does not match the submitted form data");
+  }
+
+  // Verify signature against the message
   const valid = await verifyMessage({
     address,
     message,
```
pedrohasf, line 49: One nice little improvement is instead of verifying the message we could verify the expectedMessage instead and remove the if that does `message !== expectedMessage`

non-blocking and just a nit

pedrohasf (dismissed):

PR #1173, opened 2025-09-04
feat: ENS reverse lookup in proposal

atomauro:
## Summary
- Added IntersectionObserver-based lazy loading (enabled by default).
- Wired useEnsName with query.enabled to defer network calls until in-view.
- Keeps a truncated address as the initial value; swaps to ENS if found.
- Returns a <span> element with the resolved text.

### DIFF
--- a/src/components/shared/ENSName.tsx
+++ b/src/components/shared/ENSName.tsx
@@ -1,34 +1,71 @@
 "use client";
 
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { useEnsName } from "wagmi";
 import { truncateAddress } from "@/app/lib/utils/text";
 
 interface ENSNameProps {
   address: string;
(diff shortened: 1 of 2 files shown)


src/components/shared/ENSName.tsx
```diff
@@ -1,34 +1,71 @@
...
+  const [isInView, setIsInView] = useState(!lazy);
+  const ref = useRef<HTMLSpanElement | null>(null);
+
+  // Observe visibility to enable the ENS query lazily
+  useEffect(() => {
+    if (!lazy) return;
+    const node = ref.current;
+    if (!node) return;
+
+    const observer = new IntersectionObserver(
```
sudheerDev, line 32: Hey. Why do we need IntersectionObserver to solve this?

jefag:
The domain didn't have the primary address set.

PR #1183, opened 2025-09-15
OPAG-231: hide exclude abstain option from proposal creation

aarna-arora:


pedrohasf (approved):
I would say we can delete it instead of just commenting it out.

But LGTM in whatever way.

PR #1185, opened 2025-09-16
ENG-2114/add duna-description flag for uniswap

aarna-arora:


src/app/duna/components/DunaAdministration.tsx
```diff
@@ -36,6 +37,14 @@ const DunaAdministration = async () => {
...
         </h3>
       </div>
+      {(function () {
+        const toggle = Tenant.current().ui.toggle("duna-description");
+        const content = (toggle?.config as { content?: React.ReactNode })
+          ?.content;
+        return toggle?.enabled && content ? (
+          <p className="text-sm text-primary mb-6">{content}</p>
+        ) : null;
+      })()}
```
pedrohasf, line 47: In general i avoid IIFEs as much as possible.

Instead here we could create a component similar to this:

```
const DunaDescription = () => {
  const { ui } = Tenant.current();
  const toggle = ui.toggle("duna-description");
  const content = (toggle?.config as UIDunaDescriptionConfig)?.content;
  
  if (!toggle?.enabled || !content) {
    return null;
  }

  return <p className="text-sm text-primary mb-6">{content}</p>;
}; 
```

And use it instead

pedrohasf (commented):
Small change request and we should be good to go!

pedrohasf (approved):

PR #1189, opened 2025-09-17
Aggregate delegate statements by DAO in `getVotes` query

Daktic:
Refactored code to improve clarity and functionality by aggregating delegate statements by DAO slug and address within the `getVotes` query. Adjusted log levels to reduce noise in the console during runtime.

### DIFF
--- a/src/app/api/common/votes/getVotes.ts
+++ b/src/app/api/common/votes/getVotes.ts
@@ -348,19 +348,19 @@ async function getVotersWhoHaveNotVotedForProposal({
         ),
             -- Aggregate delegate statements by address and DAO slug, keeping the latest updated_at
             aggregated_delegate_statements AS (
-          SELECT
-          address,
-          dao_slug,
-          MAX(twitter) AS twitter,  
-          MAX(discord) AS discord,        
-          MAX(warpcast) AS warpcast,            
-          MAX(updated_at) AS latest_updated_at  
+          SELECT DISTINCT ON (address, dao_slug)
+            address,
+            dao_slug,
+            twitter,
+            discord,
+            warpcast,
+            updated_at AS latest_updated_at
           FROM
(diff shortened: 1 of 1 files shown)


src/app/api/common/votes/getVotes.ts
```diff
@@ -346,6 +346,22 @@ async function getVotersWhoHaveNotVotedForProposal({
...
+          MAX(twitter) AS twitter,  
+          MAX(discord) AS discord,        
+          MAX(warpcast) AS warpcast,            
+          MAX(updated_at) AS latest_updated_at  
+          FROM
+          agora.delegate_statements
+          WHERE
+          dao_slug = $6::config.dao_slug
+          GROUP BY
+          address, dao_slug -- Ensures one row per address and DAO slug
```
jefag, line 363: MAX(twitter) is not “the latest twitter handle” — it’s just the lexicographically greatest non-NULL value among all rows for that (address, dao_slug).

If someone removes their Twitter handle (by setting it to NULL in a new row), that row will still get ignored by MAX(), because MAX() skips NULLs.

That means if they ever had a Twitter handle in the past, the query will keep returning it forever, because the max of [NULL, "oldhandle"] is "oldhandle".

Daktic: Removed max calls in favor of a select distinct ordered desc on updated_at

jefag: While this second patch might work, it's not fixing the root cause of the problem and it's slowing down the query because it has to do a distinct on those two columns.

The right thing to do here, is to track down WHY the 2nd row is being returned at all.  Like, is it the fact that his delegate statement is borked somehow?  Or does his ENS name point at two addresses? Or maybe there is a problem in our indexing layer, where we have a duplicate record for some reason.

PR #1193, opened 2025-09-19
fix(bug): “Hasn’t voted” list

atomauro:


jefag:
@atomauro Thanks for this, was there a ticket for this?

PR #1201, opened 2025-09-29
feat(security): atom/ENG-2022/Improve security proposal drafts - Sign In With Ethereum (SIWE)

atomauro:
### Internal Document

`SIWE-Security-for-Draft-Proposal-Endpoints` (reference): [Notion](https://www.notion.so/agoraxyz/SIWE-Security-for-Draft-Proposal-Endpoints-262528b28fc4818d9a14f1b802f1236c)

### Summary

- **Scope**: This PR is SIWE-only for drafts.
- **UUID**: Already resolved in `7d2f0b11` (UUID-only). No 301 or dual lookup here.
- **Security**: All draft reads/mutations require SIWE (JWT) + ownership (`author_address`).
- **Signatures**: Supports EOA (EIP-191) and multisig/SCW (EIP-1271) with robust on-chain verification.
- **UX**: Explicit CTA “Sign access request”, clear states (Awaiting Confirmation / Signed / Cancelled), route-level skeleton, and deterministic progress once JWT exists.

### Changes

- **AuthN/AuthZ**

  - New protected GET `GET /api/v1/drafts/[id]` (UUID-only): requires JWT and validates ownership.
  - `/api/v1/auth/verify`:
    - Parses client SIWE EIP-4361.
    - EOA verification (`verifyMessage`) and 1271 fallback (`isValidSignature(hashMessage(msg), sig)`).
    - Issues JWT; tolerates absence of `badgeholders` in dev.
  - Server actions (create/edit/delete/sponsor): require SIWE and validate ownership before mutating.

- **UI/UX**

  - `DraftProposalPageClient` with `@tanstack/react-query`:
    - `queryKey: ["draft", idParam, stageIndex]`, `refetchOnMount: "always"`, `staleTime: 0`.
    - Subcomponents: `useDraftStage`, `DraftPageHeader`, `SiweAccessCard`, `ForbiddenAccessCard`.
    - Deterministic progress: poller advances once JWT exists and no draft is loaded (no timeouts).
    - Smooth handling: 401 → SIWE CTA (without purging session); 403 → “Access denied” with “Back to Home”; cancellation → “Cancelled”.
  - No global auto-SIWE: signing only within the draft flow.

- **Infra/utilities**
  - 1271 fallback uses `getPublicClient()`; compared with centralized `EIP1271_MAGIC_VALUE` in `src/lib/constants.ts`.
  - `FormState` centralized in `src/app/types.ts`; `revalidatePath` returns `FormState`.
  - Cleaned Drafts links to UUID-only in UI (without `uuid ?? id`).

### QA

- Open draft by UUID without JWT → CTA “Sign access request” → sign → load draft; edit/delete work.
- Open by numeric ID → no redirect; considered invalid/Not Found.
- Multisig (Safe) works via 1271; if signer is not owner → “Access denied”.
- No auto-SIWE outside drafts.
- Multi-tenant: signing/verification on the tenant's governor chain.

### Chain Usage in SIWE

- **Frontend (signing)**: chain from `Tenant.current().contracts.governor.chain.id`.
- **Backend (verify)**:
  - EOA: `verifyMessage` (chain-agnostic).
  - 1271: tenant's `getPublicClient()` → `isValidSignature(hashMessage(message), signature)` → compares with `0x1626ba7e`.

### Environment Variables

- **`JWT_SECRET`**: secret for signing SIWE JWTs (required by `/api/v1/auth/verify`).

### DIFF
--- a/src/app/proposals/draft/[id]/DraftProposalPageClient.tsx
+++ b/src/app/proposals/draft/[id]/DraftProposalPageClient.tsx
@@ -53,8 +53,8 @@ export default function DraftProposalPageClient({
   const { switchChainAsync, isPending: isSwitching } = useSwitchChain();
 
   useEffect(() => {
+    // Do not clear SIWE session on disconnect/refresh; only reset local UI state
     if (!address) {
-      // Do not clear SIWE session on disconnect/refresh; only reset local UI state
       setDraft(null);
       hasDraftRef.current = false;
       setIsSigning(false);
@@ -65,16 +65,11 @@ export default function DraftProposalPageClient({
         clearInterval(pollIdRef.current);
         pollIdRef.current = null;
       }
-    }
-  }, [address]);
-
-  // Preserve existing SIWE session across reconnect/refresh
-  useEffect(() => {
-    if (address) {
-      wasConnectedRef.current = true;
-    } else {
       wasConnectedRef.current = false;
+      return;
     }
+    // Connected
+    wasConnectedRef.current = true;
   }, [address]);
 
   const {
@@ -153,25 +148,7 @@ export default function DraftProposalPageClient({
       }
     } catch {}
   }, []);
-  useEffect(() => {
-    if (!isSigning) return;
-    const id = setInterval(() => {
-      try {
-        const stage = localStorage.getItem(LOCAL_STORAGE_SIWE_STAGE_KEY);
-        if (completedSignRef.current) return;
-        if (stage === "awaiting_response") {
-          setSignLabel("Awaiting Confirmation");
-        } else if (stage === "signed") {
-          completedSignRef.current = true;
-          setSignLabel("Signed");
-        } else if (stage === "error") {
-          setSignLabel("Cancelled");
-        }
-      } catch {}
-    }, 200);
-    pollIdRef.current = id;
-    return () => clearInterval(id);
-  }, [isSigning]);
+  // consolidated into main polling effect
 
   const handleSiwe = useCallback(async () => {
     try {
@@ -324,6 +301,7 @@ export default function DraftProposalPageClient({
         }
       } catch {}
     }, 250);
+    pollIdRef.current = id;
     return () => clearInterval(id);
   }, [isSigning, signLabel, draft, error, refetch]);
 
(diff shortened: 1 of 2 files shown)


sudheerDev (changes requested):

src/components/shared/SiweProviderConfig.tsx
```diff
@@ -18,13 +23,13 @@ export const AGORA_SIGN_IN_MESSAGE = "Sign in to Agora with Ethereum";
 // JWT tokens for SIWE should therefore be issued with a short expiry time.
 */
 
-const isSiweEnabled = () => {
-  return process.env.NEXT_PUBLIC_SIWE_ENABLED === "true";
-};
+const isSiweEnabled = () => false; // Disable SIWE by default; no auto-trigger
```
pedrohasf: We don't need this function if it's always the same

src/app/proposals/draft/[id]/DraftProposalPageClient.tsx
```diff
@@ -0,0 +1,411 @@
...
+      <div className="max-w-screen-xl mx-auto mt-10">
+        <SiweAccessCard
+          error={null}
+          isSigning={isSigning}
+          signLabel={signLabel}
+          onConnectClick={() => {}}
+          onSignClick={handleSiwe}
+          isSwitching={isSwitching}
+          hasAddress={Boolean(address)}
+        />
```
Comment (pedrohasf) [inline src/app/proposals/draft/[id]/DraftProposalPageClient.tsx]: Both usages of this component is defining `onConnectClick` as an empty function. We can probably remove it.

src/app/proposals/draft/[id]/DraftProposalPageClient.tsx
Comment (pedrohasf) [inline src/app/proposals/draft/[id]/DraftProposalPageClient.tsx]: Slightly concerned by how many `useEffect` exist on this file. Feels like a likely source of bugs. There are some things we could do to improve here:

1. Merge effects 1 and 2:  both depend on address and can be combined.

2. Remove effect 3: don’t sync react-query state to local state; use `data`, `isLoading`, and `queryError` directly.

3. Consolidate polling effects: they overlap.

4. Move the large effect into a custom hook.

pedrohasf (changes requested):
Nothing big. Just some nits mostly. Still important to clean up though.

The comment

> If createProposalDraft throws, isPending never resets and the button gets stuck in loading state.

Happens many times throughout the changes. I picked just one, please take a look at this pattern.

src/app/proposals/draft/[id]/DraftProposalPageClient.tsx
Comment (atomauro) [reply src/app/proposals/draft/[id]/DraftProposalPageClient.tsx]: Improved. Now it relies on react-query’s values instead of syncing to local state.

PR #1202, opened 2025-09-29
fix(scope): Show friendly scopes rejection message on submit

atomauro:
Map unknown revert signatures (e.g., 0x31f837ca) and decode errors to a human-friendly scopes message in parseError Gate by Tenant.current().contracts.supportScopes to avoid false positives Surface the message in Approval and Optimistic submit UIs (use parseError); Basic already used it No behavior change for tenants without scopes.

### DIFF
--- a/src/app/proposals/draft/utils/stages.ts
+++ b/src/app/proposals/draft/utils/stages.ts
@@ -125,5 +125,34 @@ export const parseError = (error: any) => {
   if (error.message.includes("one live proposal per proposer")) {
     return "You have an outstanding proposal that has not yet completed. Please wait for it to be processed before submitting or sponsoring a new one. ";
   }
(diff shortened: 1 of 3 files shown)


src/app/proposals/draft/utils/stages.ts
```diff
@@ -125,5 +125,34 @@ export const parseError = (error: any) => {
...
+  const combined = `${msg}\n${cause}`;
+
+  const signaturesToFlag = ["0x31f837ca"];
+
+  const indicativePhrases = [
+    "unable to decode signature",
+    "decodeerrorresult",
+    "invalid proposal type",
+    "scoped",
+  ];
```
sudheerDev, line 141: Are all these errors indicative of scope issues?

PR #1448, opened 2026-04-02
Fix forum stored XSS rendering

galo13eth:


atomauro (approved):
LGTM! The `sanitize-html `implementation is solid and the tests cover the necessary edge cases perfectly.

Minor non-blocking note: Your custom `decodeHtmlEntities` works great for basic entities. If we ever encounter numerical or hex-encoded HTML entities from the forum backend in the future, we might want to consider swapping it for a robust package like `he` to prevent display logic issues, as they won't be decoded right now. Safe to merge though!

PR #1492, opened 2026-04-27
pg pretty view

aarna-arora:


src/lib/knownSelectors.tsx
```diff
@@ -858,6 +907,286 @@ export const KNOWN_SELECTORS: Record<string, SelectorAdapter> = {
...
+              {members.map((addr, i) => (
+                <div key={i}>
+                  <a
+                    href={`https://etherscan.io/address/${addr}`}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    className="font-mono text-xs hover:underline"
+                  >
+                    {addr}
+                  </a>
```
pedrohasf, line 1045: Would be nice to use the function `getBlockScanAddress` that exists in `src/lib/utils.ts` to get the correct block explorer relative to the tenant's chain instead of hardcoding to mainnet block explorer

pedrohasf (dismissed):

pedrohasf (dismissed):

jefag (approved):
Nice!

PR #1514, opened 2026-05-05
Fix sponsored delegation relay completion

galo13eth:
## Summary

This PR fixes the sponsored delegation relay flow after investigating production Mirador traces where users attempted governance delegation through `delegateBySig` and the gas relay.

The core issue was that some relayed delegation transactions were being submitted with a gas limit too close to the actual gas needed by the token contract. Successful transactions were using roughly 99% of their gas limit, and several failed transactions consumed the full gas limit and reverted out-of-gas. Separately, the frontend treated the sponsored flow as complete once the backend returned a relay transaction hash, instead of waiting for the transaction receipt and verifying that it actually succeeded on-chain.

## Investigation Findings

The broken traces were sponsored governance delegation flows on Ethereum mainnet. They all used the same relay / sponsor address:

`0x9b4B7Bd2c83B4116c2F08D2B01aE56FA55e761E1`

Three relayed transactions landed on-chain and failed by consuming the entire gas limit:

- `prod-539`: https://etherscan.io/tx/0x9580101064fd826be6cb4266ad2fc3b47209f85bf36f69bffb918c476e88d158
- `prod-569`: https://etherscan.io/tx/0x0906bb7107ce51ebc8fb52e2b495365f4c0ccf1e52302e3f079233584a06f950
- `prod-606`: https://etherscan.io/tx/0x04e7abc3e0405838ed33fcd0d0347ae5050851c20eb8b5ab27b8aa946ae9a2b9

The other failed traces in the same group timed out before we found a mined transaction, but they followed the same sponsored delegation flow pattern.

Important behavior from the investigation:

- successful relayed `delegateBySig` transactions were very close to their gas limit
- failed relayed transactions consumed the entire gas limit
- the frontend marked sponsored delegation as completed after relay submission, not after receipt success
- Mirador traces could therefore look like completed app flows even when the relayed transaction later failed on-chain

### Add a generic gas buffer for relayed `delegateBySig`

The relay backend now estimates gas for the `delegateBySig` call and submits the sponsored transaction with a generic 30% buffer over the estimate.

This is intentionally generic:

- no Ethereum-mainnet-specific gas floor
- no tenant-specific or token-specific hardcoded minimum
- no special casing for the Uniswap Foundation flow
- the gas limit is still based on the actual contract call estimate

### Wait for sponsored transaction receipts in the frontend

`useSponsoredDelegation` now waits for the relayed transaction receipt after the backend returns the tx hash.

The sponsored flow is only marked successful after:

- the relay endpoint returns a transaction hash
- the transaction is mined
- the receipt status is `success`

If the receipt reverts, times out, or relay submission fails, the hook records an error and the dialog shows the retry state.

### Avoid duplicate receipt tracking in delegation dialogs

The delegation and undelegation dialogs no longer pass sponsored relay transaction hashes into their local `useWaitForTransactionReceipt` watchers.

Direct wallet transactions are still handled by the dialog-level receipt watcher. Sponsored relay transactions are now handled inside `useSponsoredDelegation`, because that hook owns the relay request and has the right context to distinguish submitted, succeeded, and failed sponsored flows.

### Improve Mirador trace accuracy

Sponsored delegation traces now represent the actual outcome more accurately:

- relay submission event when the backend returns a transaction hash
- success close only after a successful on-chain receipt
- failure close when relay submission fails, the receipt reverts, or the receipt wait times out
- tx hints keep readable sponsored delegation / undelegation labels

This should make future Mirador traces distinguish between user exit, relay submission failure, relayed transaction failure, and true delegation success.

## Validation

Added focused coverage for `useSponsoredDelegation`:

- successful sponsored delegation is only marked fetched after a successful receipt
- reverted sponsored delegation receipt marks the hook as errored and closes the Mirador trace as failed

Ran:

```sh
pnpm exec vitest run src/hooks/__tests__/useSponsoredDelegation.test.tsx
```

### DIFF
--- a/src/app/api/v1/relay/delegate/delegate.ts
+++ b/src/app/api/v1/relay/delegate/delegate.ts
@@ -11,6 +11,13 @@ import { createWalletClient, parseSignature, isHex } from "viem";
 import { privateKeyToAccount } from "viem/accounts";
 
 const SPONSOR_PRIVATE_KEY = process.env.GAS_SPONSOR_PK;
+const DELEGATE_BY_SIG_GAS_BUFFER_PERCENT = 30n;
+
+function getBufferedDelegateBySigGasLimit(estimatedGas: bigint) {
+  return (
+    estimatedGas + (estimatedGas * DELEGATE_BY_SIG_GAS_BUFFER_PERCENT) / 100n
+  );
+}
 
 export async function delegateBySignatureApi({
   signature,
@@ -135,14 +142,22 @@ async function prepareDelegateBySignatureApi({
   }
 
   const account = privateKeyToAccount(SPONSOR_PRIVATE_KEY);
-
-  const { request } = await publicClient.simulateContract({
+  const delegateBySigRequest = {
     address: token.address as `0x${string}`,
     abi: token.abi,
     functionName: "delegateBySig",
     args: [delegatee, BigInt(nonce), BigInt(expiry), v, r, s],
     account: account,
-  });
+  } as const;
 
-  return { request };
+  const { request } = await publicClient.simulateContract(delegateBySigRequest);
+  const estimatedGas =
+    await publicClient.estimateContractGas(delegateBySigRequest);
+
+  return {
+    request: {
+      ...request,
+      gas: getBufferedDelegateBySigGasLimit(estimatedGas),
+    },
+  };
 }
--- a/src/components/Dialogs/DelegateDialog/DelegateDialog.tsx
+++ b/src/components/Dialogs/DelegateDialog/DelegateDialog.tsx
@@ -99,6 +99,7 @@ export function DelegateDialog({
     call,
     isFetching: isProcessingSponsoredDelegation,
     isFetched: didProcessSponsoredDelegation,
+    isError: didFailSponsoredDelegation,
     txHash: sponsoredTxnHash,
   } = useSponsoredDelegation({
     address: accountAddress,
@@ -121,9 +122,7 @@ export function DelegateDialog({
     isSuccess: didProcessDelegation,
     isError: didFailDelegation,
   } = useWaitForTransactionReceipt({
-    hash: isGasRelayLive
-      ? sponsoredTxnHash
-      : (localDelegateTxHash ?? delegateTxHash),
+    hash: isGasRelayLive ? undefined : (localDelegateTxHash ?? delegateTxHash),
   });
 
   const fetchData = async () => {
@@ -136,7 +135,7 @@ export function DelegateDialog({
   };
 
   useEffect(() => {
-    if (didProcessDelegation) {
+    if (didProcessDelegation || didProcessSponsoredDelegation) {
       trackEvent({
         event_name: ANALYTICS_EVENT_NAMES.DELEGATE,
         event_data: {
@@ -159,7 +158,7 @@ export function DelegateDialog({
         });
       }
     }
-  }, [didProcessDelegation]);
+  }, [didProcessDelegation, didProcessSponsoredDelegation]);
 
   useEffect(() => {
     if (isGasRelayLive || !delegationTraceRef.current) {
@@ -341,7 +340,7 @@ export function DelegateDialog({
       );
     }
 
-    if (isError || didFailDelegation) {
+    if (isError || didFailDelegation || didFailSponsoredDelegation) {
       return (
         <Button disabled={false} onClick={executeDelegate}>
           Delegation failed - try again
@@ -377,7 +376,9 @@ export function DelegateDialog({
     ) {
       fetchData();
     }
+  }, [accountAddress, tokenBalance]);
 
+  useEffect(() => {
     if (didProcessDelegation || didProcessSponsoredDelegation) {
       // Refresh delegation
       if (tokenBalance !== undefined && tokenBalance > 0n) {
@@ -387,7 +388,7 @@ export function DelegateDialog({
         });
       }
     }
-  }, [didProcessDelegation, delegate, tokenBalance, accountAddress]);
+  }, [didProcessDelegation, didProcessSponsoredDelegation]);
 
   if (!isReady) {
     return (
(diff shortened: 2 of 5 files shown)


pedrohasf (commented):
The changes for the gas limit make sense to me but the changes around useWaitForTransactionReceipt are confusing.

It shouldn't be happening and doesn't look directionally correct IMO.

If it is true then delegations without gas sponsoring would also be failing on that same path because it's the same exact code path.

The code logic is still the same, it uses the same hook, it's waiting in the same way. It just changed places.

I would expect receipt.status !== "success" to be the same thing as isSuccess which is what we use throughout the codebase. https://wagmi.sh/react/api/hooks/useWaitForTransactionReceipt#iserror-ispending-issuccess

If there is a problem here then we have a bigger problem throughout the whole codebase which seems very sus to me

galo13eth:
> The changes for the gas limit make sense to me but the changes around useWaitForTransactionReceipt are confusing.
> 
> It shouldn't be happening and doesn't look directionally correct IMO.
> 
> If it is true then delegations without gas sponsoring would also be failing on that same path because it's the same exact code path.
> 
> The code logic is still the same, it uses the same hook, it's waiting in the same way. It just changed places.
> 
> I would expect receipt.status !== "success" to be the same thing as isSuccess which is what we use throughout the codebase. https://wagmi.sh/react/api/hooks/useWaitForTransactionReceipt#iserror-ispending-issuccess
> 
> If there is a problem here then we have a bigger problem throughout the whole codebase which seems very sus to me

I agree that wagmi isSuccess already means the receipt wait succeeded. The issue is not wagmi missing reverted receipts.

The difference is that sponsored delegation had a second success flag (`didProcessSponsoredDelegation`). useSponsoredDelegation was setting `isFetched=true` immediately after the relay API returned a tx hash. That is earlier than receipt success. Non-sponsored delegations only have the wagmi receipt as a success source, so they don't have this dual-flag issue.

So it was not the exact same code path as non-sponsored delegation. The dialog did also watch the tx receipt, but the sponsored hook itself exposed relay submission as success. This PR changes that semantic so sponsored isFetched only happens after the receipt succeeds.

Please lmk if I'm reading this incorrectly.

pedrohasf:
> The difference is that sponsored delegation had a second success flag (`didProcessSponsoredDelegation`). useSponsoredDelegation was setting `isFetched=true` immediately after the relay API returned a tx hash. That is earlier than receipt success. Non-sponsored delegations only have the wagmi receipt as a success source, so they don't have this dual-flag issue.
> 
> So it was not the exact same code path as non-sponsored delegation. The dialog did also watch the tx receipt, but the sponsored hook itself exposed relay submission as success. This PR changes that semantic so sponsored isFetched only happens after the receipt succeeds.
> 
> Please lmk if I'm reading this incorrectly.

Thank you, this makes more sense.

So:

Sponsored:
```
const {
    call,
    isFetching: isProcessingSponsoredDelegation,
    isFetched: didProcessSponsoredDelegation,
    txHash: sponsoredTxnHash,
  } = useSponsoredDelegation({
    address: accountAddress,
    delegate,
  });
```

vs

Non sponsored:
```
const {
    isLoading: isProcessingDelegation,
    isSuccess: didProcessDelegation,
    isError: didFailDelegation,
  } = useWaitForTransactionReceipt({
    hash: isGasRelayLive
      ? sponsoredTxnHash
      : (localDelegateTxHash ?? delegateTxHash),
  });

```

So it seems the root of the problem is actually the flag renaming, right?  `isFetched: didProcessSponsoredDelegation,` vs     `isSuccess: didProcessDelegation,`

Agreed that this looks incorrect. We should use `isSuccess` for sponsored flow as well.

Is there any reason why we can't make this one line change? Instead of the much bigger change we currently have in the PR

galo13eth:
Yeah, the naming is definitely part of what makes this confusing, but I do not think a rename alone fixes it.

The issue is not just that `isFetched` got aliased to `didProcessSponsoredDelegation`. The issue is when that flag becomes true.

In the original sponsored hook we had:

```ts
const hash = await response.json();

// attach tx hash / close trace as submitted

setTxHash(hash);
setIsFetched(true);
```

At that point the relay endpoint has only returned a tx hash. It has not waited for the tx to be mined. So `isFetched=true` meant "relay submitted and gave us a hash", not "delegation succeeded on-chain".

That is why a one-line rename would still be wrong. It would just rename the early submitted state to success.

The dialog then had two completion flags:

```ts
isFetched: didProcessSponsoredDelegation, // from useSponsoredDelegation
isSuccess: didProcessDelegation,          // from useWaitForTransactionReceipt
```

And some completion logic accepted either:

```ts
if (didProcessDelegation || didProcessSponsoredDelegation) {
  // refresh / show completed
}
```

We could make a smaller change than the current PR by keeping the receipt hook in the dialog and only using `useWaitForTransactionReceipt().isSuccess` for final completion. I think that can work, but it has a downside.

The tradeoff is that loading/error/Mirador state then stays split across two places. The reason I moved the receipt wait into `useSponsoredDelegation` is that the hook already owns the sponsored flow: signing, relay call, tx hash, loading/error state, and Mirador trace. With the wait inside the hook, the state is easier to understand:

```ts
isFetching = signing / relay / receipt wait is still in progress
isFetched = sponsored tx succeeded on-chain
isError = signing failed, relay failed, receipt reverted, or receipt timed out
```

pedrohasf:
Ok, fair, makes sense.

Tbh i would rather make a smaller change here, i get the tradeoff of

> The tradeoff is that loading/error/Mirador state then stays split across two places

But it's just a question of what we split because right now the PR is creating a split between where the `useWaitForTransactionReceipt` for sponsored vs non sponsored happens.

And that's core functionality while Mirador isn't. So i would rather split Mirador then split `useWaitForTransactionReceipt` 

So, it's a decision of what we split.

In anyway this makes sense to me now and i don't think it really matters that much, approving the PR

pedrohasf (approved):

PR #1522, opened 2026-05-12
Lucas/fix vote history infinite scroll

galo13eth:


src/app/api/proposals/getVotesChart.ts
```diff
@@ -1,6 +1,9 @@
 import Tenant from "@/lib/tenant/tenant";
 import { TENANT_NAMESPACES } from "@/lib/constants";
-import { SnapshotVotePayload, VotePayload } from "@/app/api/common/votes/vote";
+import type {
+  SnapshotVotePayload,
+  VotePayload,
+} from "@/app/api/common/votes/vote";
```
sudheerDev, line 6: Do we need these type changes?

src/components/duna-editor/DunaContentRenderer.tsx
```diff
@@ -164,6 +162,7 @@ export default function DunaContentRenderer({
   );
   const isMarkdown = !looksLikeHtml(sanitizedContent);
   const renderedContent = useMemo(() => {
+    if (!content) return null;
```
sudheerDev, line 165: Unintended change?

src/components/ui/tooltip.tsx
```diff
@@ -15,15 +15,18 @@ const TooltipContent = React.forwardRef<
...
-  <TooltipPrimitive.Content
-    ref={ref}
-    sideOffset={sideOffset}
-    className={cn(
-      "z-50 overflow-hidden rounded-md border bg-neutral px-3 py-1.5 text-sm text-secondary shadow-md animate-in fade-in-0 zoom-in-95 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2 cursor-auto",
-      className
-    )}
-    {...props}
-  />
+  <TooltipPrimitive.Portal>
```
sudheerDev, line 18: Is this an intentional change?

sudheerDev (dismissed):
Can we remove any change that is not needed from this PR?

src/app/api/proposals/getVotesChart.ts
galo13eth: This is a Mauricio fix. But yeah, I believe we do need this fix. We are importing from a .d.ts declaration file. So we should import them as types to avoid crashing the page on SSR.

src/components/duna-editor/DunaContentRenderer.tsx
galo13eth: Again a Mauricio fix. But yeah, seems a good addition. Won't break anything if there is content. Will prevent breaks if called with undefined/empty.

src/components/ui/tooltip.tsx
galo13eth: Yup. This one`s mine, and it's a side effect of virtualization. Virtualized rows use transform: translateY(...), which makes each row a containing block for absolute-positioned descendants. Radix's Tooltip.Content
now anchors to the row instead of an ancestor outside the scroll container, so overflow-y-auto clips it. Portal escapes that.

sudheerDev (approved):

PR #1538, opened 2026-05-26
Fix Mirador wallet tracing and voting false positives

galo13eth:
## Summary

This PR collects the fixes and trace enhancements from the Mirador production trace investigation. The main goals are to reduce false-positive Mirador errors, preserve enough wallet/provider context to debug real failures, and keep vote/delegation user flows from being blocked by observability or analytics work.

## Trace Findings Addressed

- User-cancelled wallet actions were being logged as Mirador error events because `_failed` event names always mapped to `ErrorEvent`. This created noisy false positives for normal user behavior such as rejected signatures, rejected transactions, Privy link/update exits, and provider `4001` cancellations.
- Some wallet traces showed `n.connector.getChainId is not a function`. The likely scenario was a stale/serialized Wagmi connector object being used before the live connector methods were available, especially around reconnecting or hydration timing.
- Sponsored vote relay traces could stop after broadcast without enough context to tell whether the transaction was confirmed, timed out locally, failed on-chain, or was rejected because the voter had already voted.
- Direct delegation and undelegation failures were closing the trace as soon as Wagmi exposed a failure flag, before the actual write/receipt error object was available. That collapsed useful errors into generic `Delegation transaction failed` / `Undelegation transaction failed` messages.
- Direct delegation via Viem could fall back to Wagmi even when the first error was a user cancellation, creating extra work and confusing trace output for a user-initiated cancel.
- Vote analytics awaited on the voting success path. Analytics should be best-effort and should not delay or break the voting flow.
- SIWE nonce generation was worth making explicitly dynamic/no-store because one investigated trace hit an expired/unknown nonce path.
- Dismissing the SIWE sign-in prompt (ConnectKit `doSignIn()` returning `false`) was logged as `siwe_login_failed`, an error-severity event, even though closing the prompt is normal user behavior. Genuine nonce/verify failures already surface through their own events.
- A class of real `An unknown RPC error occurred` failures are wallet-side transport problems — WalletConnect relay publish failures (`Failed to publish payload`), the wallet's own RPC returning an HTTP 4xx (`RPC endpoint returned HTTP client error`), and JSON-RPC id mismatches (`Invalid Id`) — rather than app or backend outages, yet they paged at `error` severity. Separately, some traces carried a user rejection only in the nested `Details:` line under an RPC-error headline, so they should still classify as a cancellation.
- Vote and delegation failure traces flattened the wallet/provider error down to a single message string, losing the structured fields (error name/code, viem `Details:` line, HTTP status, RPC host) needed to tell whose RPC failed and to separate transport failures from real reverts.

### Mirador severity classification

- Adds `src/lib/mirador/eventSeverity.ts` as the shared classifier for Mirador web/server event severity.
- Maps recognized user-cancellation details to `info` even when the event name is a `_failed` event.
- Keeps real failures as `error` and keeps warning-style events such as `_skipped`, `_mismatch`, and `_replaced` as warnings.
- Covers common cancellation shapes and messages, including provider `4001`, `ACTION_REJECTED`, `USER_REJECTED`, `user rejected`, `user denied`, `denied transaction signature`, `rejected the request`, `exited_link_flow`, and `exited_update_flow`.
- Classifies wallet-to-chain transport failures as `warn` rather than `error`, so they stay visible without paging like a real backend outage. Covered signatures: WalletConnect relay `failed to publish payload`, viem `rpc endpoint returned http client error`, and JSON-RPC `details: invalid id`. The cancellation check runs first, so a user rejection wrapped in an RPC-error headline still resolves to `info`.
- Updates both `webTrace` and `serverTrace` to use the shared classifier.

### Wallet connector readiness and trace context

- Adds `getWalletTransactionReadinessError` to fail early with a user-actionable message when the wallet is not connected, is still reconnecting, or the connector object lacks live transaction methods such as `getAccounts` / `getChainId`.
- Extracts `checkWalletReadinessOrCloseTrace` to encapsulate the readiness check, trace close, and ref cleanup pattern used across all voting hooks.
- Adds `getWalletTraceAttributes` so voting/delegation traces capture account status, connected chain, target chain, connector identity, available connector methods, and injected provider flags.
- Applies readiness checks and wallet trace attributes to standard voting, advanced voting, sponsored voting, delegate, and undelegate flows.
- Configures explicit Wagmi connectors in `Web3Provider`, including Safe iframe support, all injected wallets, Coinbase Wallet, and WalletConnect when a project id is configured.

### Wallet error diagnostics

- Adds `getWalletErrorDiagnostics` (in `src/lib/wallet/errors.ts`) to extract structured fields from a viem/wallet error by walking the `cause` chain: error name, error code, short message, the viem `Details:` line, HTTP status, and RPC host.
- Extracts the RPC host without its query string, so the `?secret=` on RPC URLs is never sent to Mirador.
- Attaches these diagnostics to standard, advanced, and sponsored vote failures and to delegate/undelegate failures. This is what lets transport failures be told apart from real reverts and lets the failing RPC be identified for triage.

### Sponsored vote relay handling

- Adds relay vote client-error classification for `voter already voted`, returning a 409 with a stable `VOTER_ALREADY_VOTED` code instead of a generic 500.
- Stops leaking internal error details in the relay vote 500 response body — the generic fallback now returns `"Internal server error"` only, with full details preserved in the Mirador server trace.
- Enhances server-side relay trace events with sponsor address, governor address, chain id, sponsor latest/pending nonce snapshots, gas fields, request nonce, and broadcast state.
- Renames/records the successful write stage as `relay_vote_broadcasted` with explicit `confirmationState: unconfirmed` rather than implying final success at broadcast time.
- Adds client-side Mirador events for relay broadcast, relay confirmation, and relay confirmation failure/timeout.
- Adds a 10-minute sponsored vote receipt timeout and a clearer user-facing error when the relay transaction was broadcast but not confirmed locally in time.

### Delegation trace fixes

- Waits for the actual Wagmi write or receipt error before closing delegation/undelegation traces as failed.
- Preserves the real wallet/provider error message in the Mirador details instead of replacing it with a generic fallback.
- Extracts `getWalletErrorMessage` as a shared utility for delegation error messages (previously duplicated in DelegateDialog and UndelegateDialog).
- Treats Viem direct-delegation user cancellations as terminal user cancels and avoids falling through to the Wagmi fallback path.
- Adds wallet trace attributes to delegate and undelegate trace starts.

### Analytics isolation

- Converts vote analytics tracking in standard, advanced, and sponsored voting flows to fire-and-forget work.
- Extracts `trackEventFireAndForget` helper to replace repeated `void Promise.resolve().then(() => trackEvent(...)).catch(...)` patterns.
- Catches and logs analytics failures locally so analytics cannot block, delay, or break a completed vote flow.

### SIWE nonce and included support fixes

- Makes the SIWE nonce route run in the Node runtime, force dynamic behavior, and return `Cache-Control: no-store`.
- Closes a dismissed SIWE sign-in (ConnectKit `doSignIn()` returning `false`) as `siwe_login_cancelled` (info) instead of `siwe_login_failed`, using a dedicated sentinel error to distinguish a user-closed prompt from a genuine sign-in failure.
- Improves Agora API error propagation by preserving response status, status text, and response body-derived messages.


src/scripts/backfill-notification-recipients.ts
aarna-arora, line 1: this script change seems to be not mirador related - is this intentional?

aarna-arora (dismissed):

aarna-arora (approved):

